PR #254, opened 2015-01-10
[2.0] Cleanup & prepare for Laravel 5

barryvdh:
This PR keeps track of the changes for Laravel 5. Suggestions and issues can be mentioned here.

Tasks:
 - [x] Move to new config system
 - [x] Remove obsolete functions/checks
 - [ ] Make BC changes while it's possible?
 - [ ] Try using new Middleware?
 - [ ] New features?


GrahamCampbell:
Should we drop laravel 5 support from 1.x? Perhaps in a 1.9 release?

barryvdh:
Yeah I was thinking about how to best handle that. Problem is that if I remove 5.x in composer, it will keep installing 1.8.5 forever. Which isn't neceserraly bad, but I'd rather have everyone forced to 2.x.
So I could break L5 in 1.8.6 and then drop L5 support in 1.9 with composer. Downside is that it will break for current L5 users, upside is that everyone will be forced to move to 2.x..
(Or I could just leave them at 1.8.5. for as long as it works..)

GrahamCampbell:
Yeh, but current L5 users shouldn't be using L5, should we should just break it asap and get new 5.0 users onto the upcoming 2.x release.

Issue #233, opened 2014-11-28
Not working with app route prefixes.


matyunin:
If we have a prefix in app routes config:

``` php
Route::group(
    ['prefix' => '/my_prefix'],
    function () {
        Route::get('/', HomeController::class . '@showWelcome');
    }
);
```

`Debugbar` will render links as:

```
http://localhost:8080/_debugbar/assets/javascript?1416744301
```

But they should look like this:

```
http://localhost:8080/my_prefix/_debugbar/assets/javascript?1416744301
```

And in `<root>/vendor/barryvdh/laravel-debugbar/src/ServiceProvider.php` we can see this logic:

``` php
        $this->app['router']->get(
            '_debugbar/open',
            array(
                'uses' => 'Barryvdh\Debugbar\Controllers\OpenHandlerController@handle',
                'as' => 'debugbar.openhandler',
            )
        );
```

That doesn't affect main app prefix.

Please, fix this _little_ problem.

PR #263, opened 2015-01-26
Path prefix support for asset routes (for #233)

FractalizeR:
This small PR adds an ability to define URL prefix for asset routing. This solves cases like #233 (they arise, for example, if you have a website under a virtual, non-physical folder in nginx and Laravel doesn't detect any base path).

I've tested it on my vagrant configuration, which has such virtual folder configured. All seem to run fine.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -43,7 +43,7 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
         }
 
         $this->app['router']->get(
-            '_debugbar/open',
+            $this->app['config']->get('laravel-debugbar::config.path_prefix', '') . '_debugbar/open',
             array(
                 'uses' => 'Barryvdh\Debugbar\Controllers\OpenHandlerController@handle',
                 'as' => 'debugbar.openhandler',
@@ -51,7 +51,7 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
         );
 
         $this->app['router']->get(
-            '_debugbar/assets/stylesheets',
+            $this->app['config']->get('laravel-debugbar::config.path_prefix', '') . '_debugbar/assets/stylesheets',
             array(
                 'uses' => 'Barryvdh\Debugbar\Controllers\AssetController@css',
                 'as' => 'debugbar.assets.css',
@@ -59,7 +59,7 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
         );
 
         $this->app['router']->get(
-            '_debugbar/assets/javascript',
+            $this->app['config']->get('laravel-debugbar::config.path_prefix', '') . '_debugbar/assets/javascript',
             array(
                 'uses' => 'Barryvdh\Debugbar\Controllers\AssetController@js',
                 'as' => 'debugbar.assets.js',
--- a/src/config/config.php
+++ b/src/config/config.php
@@ -52,6 +52,20 @@ return array(
 
     /*
      |--------------------------------------------------------------------------
+     | Path prefix for asset routes
+     |--------------------------------------------------------------------------
+     |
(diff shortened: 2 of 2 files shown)


barryvdh:
Are you sure this fixes that? Doesn't this just add a prefix to the router, instead of the generated urls?

FractalizeR:
It does add prefix to the router, which in turn causes this prefix to be added to all generated urls, if they are generated via the router. As I can see all urls to the assets in your addon are generated via router. 

At least addon patched with this PR started working for me ;)

barryvdh:
So what is the prefix you added? It still seems weird

FractalizeR:
My local Vagrant installation has nginx, serving port 8080. Domain set to fh.local. I have different projects set up in that Vagrant machine. Each project has it's own directory in the filesystem and it's own url prefix in nginx, which is set up using ```location``` directive. An example of nginx config:

```
set $root_auth "/var/www/auth/public";

location /auth {
    root $root_auth;

    access_log  /var/log/nginx/auth_access.log main;
    error_log  /var/log/nginx/auth_error.log debug;

    try_files $uri $uri/ /auth/index.php?$query_string;

    location ~ \.php$ {
        include fastcgi_params;
        rewrite ^/auth/(.*)$ /$1 break;
        fastcgi_pass   unix:/var/run/auth.sock;
        fastcgi_param  SCRIPT_FILENAME $root_auth/index.php;
    }

    location ~ ^/auth/(.*\.(js|css|jpe?g|png|gif|ico))$ {
        alias $root_auth/$1;
    }
}
```

When I work with a project, which is served using ```/stock``` uri, all urls, constructed by Laravel doesn't have that ```/stock``` in them. So, I use something like this:

```php
Route::group(
    ['prefix' => 'stock'],
    function () {
        Route::get('/', ['as' => 'home', 'uses' => HomeController::class . '@index']);

        Route::get('orders/new', ['as' => 'order_new', 'uses' => OrderController::class . '@getNew']);
        Route::post('orders/new', ['as' => 'order_new_do', 'uses' => OrderController::class . '@postNew']);
    }
);
```

So, using this config ```order_new``` route has url of ```/stock/orders/new```. Full URL is ```http://fh.local/stock/orders/new``` in my VM

I suppose my situation is exacty situation for which Taylor implemented route prefixes for. Laravel Debug Bar doesn't have route prefixes anywhere in config so... I had to patch it somehow. It could be done with the same ```Route::group()``` call in the code of course.

barryvdh:
Okay but you could better just configure your vhost properly, so you have `stock.local` as domain or something. The prefix is applied to both the generated route, as the route that is is used.

It might work coincidentally, but that isn't by design.

barryvdh:
I actually did use a group yesterday for the L5 version: https://github.com/barryvdh/laravel-debugbar/commit/5fe00f0433990beff87a089fa3678d31bc0e7f6d
And might even make the prefix optional, but not for this reason specifically.

FractalizeR:
I didn't find out the way to make several Laravel installations work under the same domain in separate folders easily. If you just make a domain and put Laravel installation into a folder, you will have /public folder sticking out and your static files urls will need to be rewritten.

But the main reason is that on production we have a load balancer, that balances /url1 to one virtual machine and /url2 to another. So, there are no physical folders also. And I would like my test environment to closely resemble original.

FractalizeR:
> I actually did use a group yesterday for the L5 version: 5fe00f0
Yep, but that's your own prefix, not an external one.

barryvdh:
Yes I understand that, just saying that would make it a bit cleaner to change

FractalizeR:
Yep, agreed. My solution has another drawback - prefix should end with '/' or be empty. Using route group prefix there is no such problem and it's much cleaner.

PR #284, opened 2015-02-13
converts boolean variable type to string

birtanyildiz:
i am using .env file and without this code my debugbar can not load. when i use the code echo config('app.debug') in a controller code raises boolean string convertion error. after i make this change its ok :)

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -11,7 +11,7 @@ return array(
      |
      */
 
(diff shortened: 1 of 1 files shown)


barryvdh:
I dont think this is correct. A string will always evaluate to true..

birtanyildiz:
sorry i missed that. this pull request must reject :(

PR #287, opened 2015-02-15
Fix for route caching

jsmunich:
When route caching is enabled, the debugbar fails to render (Route [debugbar.assets.css] not defined.). One solution would be that routes should are still registered even when running in console environment.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -25,10 +25,6 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
             $this->app['config']->set('debugbar.enabled', false);
         }
 
-        if (! $this->app['config']->get('debugbar.enabled')) {
-            return;
-        }
-
         $routeConfig = [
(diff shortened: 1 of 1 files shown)


spaceemotion:
+1 Just wanted to post an issue for this as well.

barryvdh:
Ok thanks. Just have to check this for unwanted access etc.

barryvdh:
Thanks!

barryvdh:
Also tagged this, so should be good now in `v2.0.1` :)

juanmiguel431:
In my case, to solve the problem i had to run "php artisan route:clear" . After that, i could run again "php artisan route:cache" and DebugBar works well.

locojuhi:
i know this trend is cloded but, i tried clearing the routes a it still not showing the debugbar... any suggestion?

CindyBoy:
My debugbar is not display with the Lumen 5.5.2, too. Any suggestion?

PR #300, opened 2015-03-03
Route prefix support added for Laravel Debug Bar internal routes (Laravel 4.x)

FractalizeR:
This adds support for configuring Laravel debug bar route prefixes so that Debug bar could be used in cases, affected by bugs like this: http://trac.nginx.org/nginx/ticket/97 where root path, detected by nginx is different from the real one.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -42,29 +42,33 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
             );
         }
 
-        $this->app['router']->get(
-            '_debugbar/open',
-            array(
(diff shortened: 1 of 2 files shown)


barryvdh:
Perhaps make the default prefix `_debugbar`, just like in L5 version?

FractalizeR:
Yep, but this looked like BC so I decided to leave this as is for now. I didn't want someone to stumble over something incompatible ;)

What do you think?

barryvdh:
I definitely don't want to break BC. But there is already a prefix to all routes (`_debugbar`), only it's added manually to all routes. So that wouldn't break, right?

FractalizeR:
I think that is correct. Committed.

PR #321, opened 2015-03-28
Fix require comand line

xroot:


barryvdh:
Don't they both work?

xroot:
Don't think so . tried that but throw an error

JayBizzle:
The first one doesn't work if your using `oh-my-zsh`...at least in OS X Terminal

PR #325, opened 2015-04-02
Move enabled check higher up

fire015:
If debug bar is not enabled it should quit before the routes are registered, otherwise you can still access URL's such as /_debugbar/assets/stylesheets


D1kz:
+1

barryvdh:
This is

PR #326, opened 2015-04-04
Force left-to-right direction

Nirlah:
Fix small issues with the bar when working on a right-to-left project.


boynet:
+1 can approve this

PR #331, opened 2015-04-08
fix LogsCollector for default daily logs naming convention in L5

unitedworx:
Laravel 5 defaults to daily log files with the naming convention laravel-2015-04-08.log. updated getLogsFile() method to play nice with this change and default to the daily logs while falling back to the single file logs.

### DIFF
--- a/src/DataCollector/LogsCollector.php
+++ b/src/DataCollector/LogsCollector.php
@@ -24,14 +24,14 @@ class LogsCollector extends MessagesCollector
      */
     public function getLogsFile()
     {
-        //Default log location (single file)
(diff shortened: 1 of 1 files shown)


unitedworx:
Thanks for merging this Barry. Much appreciated

barryvdh:
And thank you for fixing it :)
I don't usually use this collector, so didn't notice any problems.

unitedworx:
No problem! Considering the number of open request i figured you were busy so thought of digging in to try and fix this myself.

Will try and dig into the view collector as well since it appears not to play well with Twig bridge and does not show the actual view data.

PR #336, opened 2015-04-21
Raplce and & or 

Tjoosten:
The `and` operator does not have the same precedence as `&&`. This could lead to unexpected behavior, use `&&` instead.

The `or` operator does not have the same precedence as `||`. This could lead to unexpected behavior, use `||` instead.


barryvdh:
Thanks, but don't think it give any problems in this situation, right?

Tjoosten:
Normally not, Why?

PR #346, opened 2015-05-18
Adding Cartalyst Sentry Auth Collector

Claymm:


barryvdh:
For what version of Sentry is this? Does that have a stable L5 release?
I'm not sure this belongs in this repository, because it's a third party package..

src/DataCollector/SentryAuthCollector.php
```diff
@@ -0,0 +1,116 @@
...
+ * Collector for Laravel's Sentry Auth provider
+ */
+class SentryAuthCollector extends AuthCollector implements Renderable
+{
+    /**
+     * @var boolean
+     */
+    protected $showName = true;
+
+    public function __construct()
```
barryvdh: Probably better to use dependancy injection and inject the real object, instead of using Facades?

PR #349, opened 2015-05-21
Refacto the way user info are displayed

Claymm:
I quickly reviewed how the user information are displayed.
Initially you were checking if the AuthIdentifier is an numeric value, then you would try to display username or email attributes.
On a specific project I don't use numeric values for my AuthIdentifier, I use serials like '003g000000Sanpv' which are non-numeric values.
What I offer here is we try to display username, email first before falling back to display the identifier so we don't check if it's numeric.


barryvdh:

PR #351, opened 2015-05-25
[L4] Fix for addon crash when DB::disableQueryLog() was called in a user application

FractalizeR:
This fixes an addon crash when DB::disableQueryLog() was used inside an application to manually turn off query logging. In this case DebugBar is unable to start with the following cryptic messages in Laravel log:

```
[2015-05-25 16:10:18] development.ERROR: Debugbar exception: Undefined property: stdClass::$id  
```

### DIFF
--- a/src/LaravelDebugBar.php
+++ b/src/LaravelDebugBar.php
@@ -262,7 +262,7 @@ class LaravelDebugbar extends DebugBar
             }
         }
 
-        if ($this->shouldCollect('db', true) and isset($this->app['db'])) {
+        if ($this->shouldCollect('db', true) and isset($this->app['db']) and $this->app['db']->logging()) {
             $db = $this->app['db'];
             if ($debugbar->hasCollector('time') && $this->app['config']->get(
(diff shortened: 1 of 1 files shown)


barryvdh:
Not sure if this method was always in L4?

FractalizeR:
You mean `disableQueryLog()`? It was at least in [4.2](https://github.com/laravel/framework/blob/4.2/src/Illuminate/Database/Connection.php#L1101)

barryvdh:
No I mean `logging()`, which isn't available in 4.0

FractalizeR:
Yep, it wasn't there in 4.0 :(

krballard:
This commit breaks the package for me personally,

barryvdh:
This one or the one that's merged for L5? What error?

krballard:
Sorry,the one that is merged with L5.. It works, just the query logging completely disappears.

barryvdh:
Reverted for now.

PR #352, opened 2015-05-25
[L5] Fix for crash when attempting to explain non-mysql queries

FractalizeR:
Currently only MySQL queries are supported for EXPLAIN so... (I work with Postgres).

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -272,7 +272,7 @@ class LaravelDebugbar extends DebugBar
                 $queryCollector->setFindSource(true);
             }
 
-            if ($this->app['config']->get('debugbar.options.db.explain.enabled')) {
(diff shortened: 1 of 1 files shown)


barryvdh:
So why not just not enable it? :P

GrahamCampbell:
Also, this doesn't work. You could be using multiple drivers.

GrahamCampbell:
We can't assume the default driver is used for eveything.

FractalizeR:
@barryvdh Well, it was enabled by default in settings :) If we support only MySQL-like EXPLANs, I think there is no need to cause an error trying to parse unparseable data. It is not really obvious that only MySQL is supported, otherwise there will be a crash.

@GrahamCampbell Good catch. Missed that. Should I fix it?

barryvdh:
It shouldn't be enabled by default: https://github.com/barryvdh/laravel-debugbar/blob/50ca2243bd38fedd2b4c9314068f7f06b79602d4/config/debugbar.php#L126

PR #353, opened 2015-05-25
[L4] Fix for crash when attempting to explain non-mysql queries

FractalizeR:
Currently only MySQL queries are supported for EXPLAIN so... (I work with Postgres).


barryvdh:
Sorry, didn't see this was for 1.8

PR #360, opened 2015-06-06
Default enable session collector.

JerseyMilker:
I find this [session collector](http://puu.sh/ieCQN/998526477d.png) very useful.  What do you think about enabling it as default collector?

If not, maybe it should be mentioned somewhere in your README as an option?


barryvdh:
Sure, thanks

PR #370, opened 2015-07-08
Add to ServiceProvider and alias fadecade for laravel 5.1

jovib:


barryvdh:
This is exactly the same, but different syntax. I'm using the new syntax now. Thanks!

PR #379, opened 2015-07-21
Delegate cache busting query string to the UrlGenerator

stanfieldr:
In our case we override the UrlGenerator to provide cache busting and a few shortcuts to mounting locations. The problem is Debugbar adds a query string to bust cache even though one exists. An easy solution would be to offload the work onto the UrlGenerator since it knows best how to hand the URL.

So something like this:
`http://focus.dev/Modules.php?modname=laravel/route?123`

Becomes:
`http://focus.dev/Modules.php?modname=laravel/route&v=123`


barryvdh:
Thanks!

PR #385, opened 2015-08-10
Bump branch alias and change min stability

GrahamCampbell:


barryvdh:
Why do I need to bump?

GrahamCampbell:
Because you've added new features so are required by semver to bump the minor version.

PR #392, opened 2015-08-25
add time to events

icfr:


barryvdh:
But this just adds the time, starting from when the collector was set up, until the actual event. So not really the duration from the event itself.

PR #394, opened 2015-09-02
Routes blacklist.

triforcely:
While using DebugBar (and enjoying it) I have stumbled on a small issue or rather feature I'd like it to have. My webapp has API part for client application and I need to return to this client a string (something like API-key). Of course handling additional data in response (like DebugBar code) does not make any sense, even in debug mode. That's why I have implemented this small feature.

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -174,11 +174,9 @@ return array(
      | DebugBar even in debug mode - e.g. plain API responses with JSON data.
      | This way API clients don't have to implement more complicated parsing.
      |
-     | Syntax supports basic routes like: "route/dostuff" and routes with parameters "route/{whateveryoulike}/dostuff
-     |
      */
     'ignored_routes' => array(
-        'retrieve/{parameterA}/{parameterB}/plain',
-        'simple/route'
+
     ),
+
 );
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -27,6 +27,7 @@ use DebugBar\Storage\RedisStorage;
 use Exception;
 
 use Illuminate\Contracts\Foundation\Application;
+use Illuminate\Http\Request as LaravelRequest;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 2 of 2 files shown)


barryvdh:
So is it text/html content type, but no html? Perhaps I can better check if the html has a closing body tag, to be stricter with changing output.

barryvdh:
Perhaps an optional strict mode for example.

src/LaravelDebugbar.php
```diff
@@ -606,6 +606,43 @@ protected function isJsonRequest($request)
...
+        {
+            if($ignored_route == "")
+                $ignored_route = "/";
+
+            $ignored_route = preg_replace("({[^{|}]+})", "[A-Za-z0-9]+", $ignored_route);
+
+            if(strlen($ignored_route) > 1 && $ignored_route[0] != "/") // To allow rules writes as "/route" and "route"
+                $ignored_route = "/".$ignored_route;
+
+            if($ignored_route[strlen($ignored_route)-1] == "/")
```
GrahamCampbell: missing braces and missing space

GrahamCampbell: for ALL places

YouCantTriforce: I made sure to use correct formatting and then used wrong one... Fortunately it can be easily fixed.

YouCantTriforce:
Do as you wish. I am going to use my branch since I don't have any other solution.

barryvdh:
I'll take a look. But it's probably cleaner to use `$request->is($pattern)`, like here:
https://github.com/laravel/framework/blob/5.1/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php#L56-L71

YouCantTriforce:
Yeah, that will be much better. Honestly, I did overlook this method.

config/debugbar.php
```diff
@@ -165,4 +165,20 @@
...
+     |
+     | There is a possibility that there are routes which should never display
+     | DebugBar even in debug mode - e.g. plain API responses with JSON data.
+     | This way API clients don't have to implement more complicated parsing.
+     |
+     | Syntax supports basic routes like: "route/dostuff" and routes with parameters "route/{whateveryoulike}/dostuff
+     |
+     */
+    'ignored_routes' => array(
+        'retrieve/{parameterA}/{parameterB}/plain',
```
barryvdh: Config doesn't need pre-defined set. And instead of parameters, better use `*`

YouCantTriforce:
I have pushed new commit which should resolve all mentioned problems.

aik099:
Maybe this can be done on route level, e.g. set a route attribute (not used during url building) and then debug bar would listen for that parameter and decide if it should enabled itself or not.

This way looking at route definition you can clearly see if debugbar would be disabled for it or not.

Your approach however allows to skip routes for which you can't change definition, like ones coming from `vendor` folder, so it's also cool.

barryvdh:
Something similar has been implemented. Thanks!

PR #399, opened 2015-09-16
Merge PR from Dukeman330/laravel-debugbar to enable standalone /debugbar route, e.g. to debug REST APIs

vlascik:


barryvdh:
So what does this do exactly?

vlascik:
You get a new route, /debugbar, where you can view info about your last request. It's useful e.g. when you're debugging single page apps or REST API calls and you don't want to mess with including debugbar into your frontend. This way it just works.

barryvdh:
So, this is just an empty page with just the debugbar, but full size? And it shows the last request bij default?

- Why the routes.php instead of adding it with the other routes?
- Why the classmap, instead of PSR-4?
- CS (spacing etc) is off on multiple places, please use PSR-2
- Why do you need a new Openhandler?
- WHy do you need a new FileStorage, can't you use `$debugbar->getStorage()`?
- Rather use Dependancy injection instead of facades.

vlascik:
Yes, and yes.
As for the other comments, well, the code is not mine, I just merged it and fixed some small bugs and compatibility with L5, didn't really have time to go through how debugbar works on the inside.
I might fix it though, when I have some more spare time, we'll see.

composer.json
```diff
@@ -16,6 +16,9 @@
         "maximebf/debugbar": "~1.10.2"
     },
     "autoload": {
+        "classmap": [
+            "src/Controllers"
```
GrahamCampbell: This is incorrect.

atehnix:
Can we expect that this functionality will be in the near future?
I would be happy to use it.

barryvdh:
No plans to implement this currently, but I'm open to a PR which fixes above listed issues.

PR #409, opened 2015-10-21
Allow for middleware to check to see if the debug bar is enabled

lukepolo:
This allows for us to not throw the debuggers exceptions 

Use Cases :

When using an API we don't want to throw a non-json exception

### DIFF
--- a/src/Middleware/Debugbar.php
+++ b/src/Middleware/Debugbar.php
@@ -41,6 +41,10 @@ class Debugbar {
      */
     public function handle($request, Closure $next)
     {
+        $enabled = function () {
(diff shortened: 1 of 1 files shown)


lukepolo:
Any update on wether this will be merged ?

barryvdh:
Isn't the middleware only added when it's enabled?

lukepolo:
Yes , true but allows us to disable on the fly

lukepolo:
In the middleware / routes

barryvdh:

kheme:
I called this link in my control, and it did it for me:
``app('debugbar')->disable();``

PR #410, opened 2015-10-24
Type-hint appropriate arguments

sebdesign:
Some methods accepting `$request` and `$response` as arguments did not have the appropriate type hints, although the dockblock tags were suggesting these classes.

This makes user mistakes hard to debug. For example:

I did a mistake in my `Exceptions\Handler.php` of returning a `View` instead of a `Response` instance.

```php
public function render(Request $request, Exception $e) {
    if ($e instanceof NotFoundHttpException) {
        return view('my.custom.404');
    }
}
```

The output was `exception 'BadMethodCallException' with message 'Method [isRedirection] does not exist on view.'`.
`LaravelDebugbar` is calling `isRedirection()` on a `$response` argument, but there has no check if that argument was indeed a `Response` instance.

By typehinting those arguments we can prevent this kind of errors and make the package less error-prone.

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -456,7 +456,7 @@ class LaravelDebugbar extends DebugBar
      * @param  \Symfony\Component\HttpFoundation\Response $response
      * @return \Symfony\Component\HttpFoundation\Response
      */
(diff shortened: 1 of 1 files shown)


barryvdh:
Thanks

PR #414, opened 2015-11-09
Disallow php < 5.5

alanmanderson:
There are several places where this packages uses the ::class attribute.
This attribute was not available before php 5.5


barryvdh:
Thanks

PR #422, opened 2015-11-30
Update ViewCollector.php for StringBladeCompiler

TerrePorter:
This change is to support my StringBladeCompiler plugin.

Without the change the debugbar generates an error, "realpath () expects parameter 1 to be a valid path, object given"

This is because the path is an object when using the StringBladeCompler plugin instead of a template file path.

Thanks.

### DIFF
--- a/src/DataCollector/ViewCollector.php
+++ b/src/DataCollector/ViewCollector.php
@@ -54,14 +54,20 @@ class ViewCollector extends TwigCollector
     {
         $name = $view->getName();
         $path = $view->getPath();
-        if ($path) {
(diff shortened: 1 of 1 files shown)


TerrePorter:
Forgot to mention this should only apply to the Laravel 5 version with StringBladeCompiler 3. The plugin is a replacement/extension to the default Laravel view.

danieliancu:
+1

beeglebug:
Is this going to be merged? I really want to use these two packages together but can't at the moment.

barryvdh:
Yes

Issue #105, opened 2014-02-27
laravel-debugbar + require.js => jQuery is not defined


dmitry-ivanov:
Hi there!

Are there any elegant solutions for `laravel-debugbar` + `require.js` usage?
Separate include (not via `require.js`) of `jquery` - helps, but this is dirty solution according to `require.js`...

Thank you!

claar:
Same problem -- debugbar is not RequireJS/AMD compatible. In looking at the source code, I'm not sure yet what the easiest way would be to fix this :/

claar:
My current workaround is to tell DebugBar not to inject its javascript:

```php
// Make Debugbar play nicely with requirejs -- tell it only to inject CSS, not JS
Debugbar::getJavascriptRenderer()->setIncludeVendors('css');
```

barryvdh:
You can also set that in the config. Or use the excludeVendor option to just remove jquery.

refear99:
@barryvdh I have the same problem. 
any solution now ?

barryvdh:
Started working on that here: https://github.com/maximebf/php-debugbar/commit/c3268ca12248485a56d8fadf6489a277ef40d2e9

Any thoughts from people with requirejs experience?

PR #423, opened 2015-12-02
Adds option for requirejs compatibility

drewrichards:
Adds an option to output the debugbar javascript in a way that is compatible with requirejs. Solves the problem noted here: https://github.com/barryvdh/laravel-debugbar/issues/105. 

It works by:
- Adding a route to a javascript file that contains all the debugbar js in AMD format
- Wrapping the generated javascript in a require call that loads the debugbar module

In addition to setting the requirejs_compatible option in config/debugbar.php it is also necessary to add a path to `debugbar-all` in the requirejs config. Locally I put javascript in `public/js` and my baseUrl is `/js` so my path entry looks like `"debugbar-all": "../_debugbar/assets/debugbar"`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": ">=5.5.0",
-        "illuminate/support": "~5.0.17|5.1.*",
+        "illuminate/support": "~5.0.17|5.1.*|5.2.*",
         "symfony/finder": "~2.6",
         "maximebf/debugbar": "~1.10.2"
     },
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -166,4 +166,17 @@ return array(
      */
     'route_prefix' => '_debugbar',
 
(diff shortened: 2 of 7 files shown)


barryvdh:
Hmm, is this something we can do upstream in https://github.com/maximebf/php-debugbar ?

igorrynkovoy:
Up.

barryvdh:
Can we do this without PHP routing in php-debugbar core? I'm interested in improving, but not really sure what the best way is.

barryvdh:
See this commit: https://github.com/maximebf/php-debugbar/commit/c3268ca12248485a56d8fadf6489a277ef40d2e9

I've added an option to render the debugbar as requirejs module, I changed it to explicitly use a name. We need to reflect these changes here, but I guess that would do the trick?

barryvdh:

PR #441, opened 2016-01-01
The current version only works in 5.2

GrahamCampbell:


barryvdh:
What do you mean? What doesn't work in 5.1?

barryvdh:
Only the try catch in the middleware, right?

GrahamCampbell:
Yeh. I guess a better fix would be to provide two middlewares.

GrahamCampbell:
Actually, no. The best fix would be to revert the catch changes.

GrahamCampbell:
Since, in 5.2, that code just never gets executed.

GrahamCampbell:
But that doesn't matter, it's not breaking 5.2, if you see what i mean?

barryvdh:
Yeah I understand, it never throws an exception any more.
I removed it because it felt a bit hacky, like it wasn't supposed to be needed. But guess you're right, 5.1 doesn't need change, 5.2 doesn't use the 'hacky' stuff, except for needed the exception handler injected, but that could also be done in the catch perhaps

Issue #438, opened 2015-12-30
Storage logs in elastic search


pavankumarkatakam:
How to use this package to store logs in elasticsearch as JSON document.

barryvdh:
This isn't meant for storing logs. Use a monolog subscriber or something.

pavankumarkatakam:
But i want all the data collected by this package specifically for laravel.

barryvdh:
It's currently stored in a Debugbar specific json file, so not sure if that's helpful for elasticsearch. How do you imagine this in elasticsearch?

pavankumarkatakam:
i just manually send the .json file to elasticsearch and viewed with kibana it working fine, But i want to use some custom storage driver in this package it automatically send the data (which is writing to json file) elastic search. If so i can visualize the response codes, time taken for each route and how many exception i got in a day all those main analytics can be derived using kibana.

pavankumarkatakam:
What is your thoughts @barryvdh . Actually Elastic and Kibana is really awesome.

pavankumarkatakam:
Closing this issue, as now there is option for custom storage.

PR #443, opened 2016-01-02
Option for Custom storage

pavankumarkatakam:
Other than file, PDO and redis, This PR add the functionality to developers to have their own Storage interface.

Working:
in config 

     'storage' => array(
        'enabled' => true,
        'driver' => 'custom', // redis, file, pdo, custom
        'path' => storage_path() . '/debugbar', // For file driver
        'connection' => null,   // Leave null for default connection (Redis/PDO)
        'provider'=>"App/Framework/ElasticStorage" // Instance of StorageInterface for custom driver
    ),

    <?php

  class ElasticStorage implements StorageInterface{
 
  }

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -31,7 +31,7 @@ return array(
         'driver' => 'file', // redis, file, pdo, custom
         'path' => storage_path() . '/debugbar', // For file driver
         'connection' => null,   // Leave null for default connection (Redis/PDO)
-        'provider'=>"" // Instance of StorageInterface for custom driver
+        'provider' => '' // Instance of StorageInterface for custom driver
     ),
 
     /*
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -28,9 +28,10 @@ return array(
      */
     'storage' => array(
         'enabled' => true,
-        'driver' => 'file', // redis, file, pdo
+        'driver' => 'file', // redis, file, pdo, custom
         'path' => storage_path() . '/debugbar', // For file driver
         'connection' => null,   // Leave null for default connection (Redis/PDO)
+        'provider'=>"" // Instance of StorageInterface for custom driver
     ),
 
     /*
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -787,7 +787,7 @@ class LaravelDebugbar extends DebugBar
                     break;
                 case 'custom':
                     $class = $config->get('debugbar.storage.provider');
-                    $storage = app($class);
+                    $storage = $this->app->make($class);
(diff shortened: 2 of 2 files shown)


pavankumarkatakam:
Here the Storage implemented class can inject services like router, storage .... etc.

config/debugbar.php
```diff
@@ -28,9 +28,10 @@
      */
     'storage' => array(
         'enabled' => true,
-        'driver' => 'file', // redis, file, pdo
+        'driver' => 'file', // redis, file, pdo, custom
         'path' => storage_path() . '/debugbar', // For file driver
         'connection' => null,   // Leave null for default connection (Redis/PDO)
+        'provider'=>"" // Instance of StorageInterface for custom driver
```
barryvdh: Can you follow CS? Spacing and quotes, like the rest of the options?

src/LaravelDebugbar.php
```diff
@@ -785,6 +785,10 @@ protected function selectStorage(DebugBar $debugbar)
                     $connection = $config->get('debugbar.storage.connection');
                     $storage = new RedisStorage($this->app['redis']->connection($connection));
                     break;
+                case 'custom':
+                    $class = $config->get('debugbar.storage.provider');
+                    $storage = app($class);
```
barryvdh: I prefer `$this->app->make($provider)`

barryvdh:
Looks interesting, but why do you need extra storage?

barryvdh:
I'm guessing for #438 then?

barryvdh:
If you have good storage created, please consider adding it via a PR to https://github.com/maximebf/php-debugbar/tree/master/src/DebugBar/Storage

pavankumarkatakam:
Yes @barryvdh that issue created by me only, Actually I need custom storage to get the control over the  data that is generated while storing so that i can achieve my requirement. I didn't say that elasticsearch is the best but for some reasons i want to use the ES. And may be some other developer want to have some storage rather than creating all the storage implementations if there is option for developers to create own storage then every developer can have their own way of storing the data and of course the file is default. So rather than integrating the ElasticStorage to package give the option for developers to develop their own.

pavankumarkatakam:
And for your information , I'm using the data generated by this package for custom analytics. Like 
1) How fast the app  is running?
2)  How many users are visiting my site?
3) What is the error rate and response code?

barryvdh:
Thanks!

pavankumarkatakam:
I think we want to mention in readme.

pavankumarkatakam:
Hi @barryvdh , may i know when you are going to release the master branch.

PR #446, opened 2016-01-16
Move appended storage path inside `storage_path` function call and indent array values

shark0der:
`storage_path` accepts a `$path`parameter and will properly append it to the storage path taking care of the directory separator itself.

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -27,11 +27,11 @@ return array(
      |
      */
     'storage' => array(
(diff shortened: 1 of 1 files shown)


barryvdh:
Thanks

shark0der:
Wow, this was fast! I was already thinking that it will take months to merge it since there are like 20 unmerged PRs.... So thank you! :smile:

Issue #436, opened 2015-12-24
Support for L5.2 multi auth


vadimsg:

barryvdh:
What do you need exactly?

vadimsg:
I have 2 auths. One for public users and one for admins. After I login in public and admin side I see in debugbar auth tab only the default guard, It doesn't switch to admin guard. 

Do you think this could be added?

barryvdh:
Probably, PR is welcome :)

PR #456, opened 2016-01-30
[5.2] Add support for multi auth

plapinski:
Laravel 5.2 supports custom user providers, but unfortunately this isn't currenty supported by Laravel Debugbar. This is my fast and dirty way to solve this problem. See also https://github.com/barryvdh/laravel-debugbar/issues/436
There are no backwards compatibility issues since original AuthCollector isn't touched. This is how it looks:

screenshot-wesele123_v2 wp 2016-01-30 03-40-58

### DIFF
--- a/src/DataCollector/MultiAuthCollector.php
+++ b/src/DataCollector/MultiAuthCollector.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Barryvdh\Debugbar\DataCollector;
+
+/**
+ * Collector for Laravel's Auth provider
+ */
+class MultiAuthCollector extends AuthCollector
+{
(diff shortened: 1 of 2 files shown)


barryvdh:
Looks good, but not so happy with the class name. Maybe MultiAuthCollector?

plapinski:
@barryvdh You're right, I renamed it.

barryvdh:
Thanks. Can you use the checkVersion() method, instead of the str_contains check?

plapinski:
Sure, done.

barryvdh:
Thanks!

plapinski:
My pleasure :)

PR #457, opened 2016-01-30
add missing use declaration for GateCollector

plapinski:
As in title, use declaration for GateCollector was missing.


barryvdh:
Thanks

PR #460, opened 2016-02-04
More pleasing display of table elements

believer-ufa:
In the Queries tab, we can see that the blocks that display time of the request and the name of the database, constantly jumped in their seats. This amendment corrects this problem


barryvdh:
Do you have a screenshot with difference? Can you submit this upstream to https://github.com/maximebf/php-debugbar

believer-ufa:
was: 

became: 

I corrected the styles of the component for laravel, and I'm not sure that they have similar problems :)

barryvdh:
Okay, will see if I move this to php-debugbar. Thanks!

Issue #247, opened 2014-12-23
No Debugbar Storage found.. when php artisan debugbar:clear


sergiodebcn:
When I execute the command 'php artisan debugbar:clear' I get an error, No Debugbar Storage found..

If I hardcode the class ClearCommand to set the storage manually it works perfect.

public function fire()
    {
/* Added Code
        $storage = new FilesystemStorage($this->debugbar->app['files'], storage_path() . '/debugbar');
        $this->debugbar->setStorage($storage);
*/
        if ($storage = $this->debugbar->getStorage()) {
            $storage->clear();
            $this->info('Debugbar Storage cleared!');
        } else {
            $this->error('No Debugbar Storage found..');
        }
    }

Thanks

tonglil:
Yes, I also have this issue, even though the /storage/debugbar folder exists with `.json` files inside.

I only stumbled upon this because I'm getting a 500 when the debugbar tries to load something like this: `https://example.com/_debugbar/open?op=get&id=1ea0b33b6eaf429b32829943a030b9da`

totalpeople:
I'm also getting this error when I run the artisan command. My env is local. I don't have any storage/debugbar folder. It's an app that I've just migrated to Laravel 5 from v4. Any ideas?

goyote:
I found the issue, in the service provider in line 62 you have this:

        if ($app->runningInConsole()) {
            $this->app['config']->set('debugbar.enabled', false);
        }

So when the following check happens `debugbar.enabled` is set to `false` and not `null`.

        // If enabled is null, set from the app.debug value
        if (is_null($enabled)) {
            $enabled = $this->app['config']->get('app.debug');
            $this->app['config']->set('debugbar.enabled', $enabled);
        }

furey:
@barryvdh I'm experiencing the same error.

-----

**Published package config**

[/config/debugbar.php](https://github.com/barryvdh/laravel-debugbar/files/33600/debugbar.php.txt)

![storage-settings](https://cloud.githubusercontent.com/assets/1914481/11138342/0bfdaf0c-8a15-11e5-8140-69a3bf871b70.png)

**Project directory (storage folder exists)**

![root](https://cloud.githubusercontent.com/assets/1914481/11138312/b89d4458-8a14-11e5-8670-621bf485b13b.png)

**Artisan output**

[...]

PR #462, opened 2016-02-10
While caching config, the enabled setting was overwritten

JaapMoolenaar:
The service provider set debugbar.enabled to false when running any console command, including config:cache. Said command uses the config array to create the cache file, and thus the configuration unexpectedly changes.

(A situation where the config is cached with the debugbar is enabled probably almost never occurs, nevertheless the debugbar is a nice performance testing tool, which includes testing the performance gain by caching the configuration)


barryvdh:
I just return without modifying the console now.

JaapMoolenaar:
That works great too (and should fix other issues as well).
I see you solved a potential issue with #247, nice job ;-)

PR #463, opened 2016-02-11
[L5] Fatal Error in Lumen 5.2.* Query Collector

Naminator:
This is the Lumen version I have. (Lumen (5.2.4) (Laravel Components 5.2.*))

Since it didn't work with the stable version of laravel-debugbar I had to download the master branch and it worked fine until I started running queries through Eloquent. All of them were crashing because LaravelDebugbar couldn't cast the event object as a string due to lack of __toString() method in the QueryExecuted object. My solution is a bit hacky and you should change it however you please. I haven't done much pull requesting before and I am not sure how you deal with this sort of things but this is the easiest solution I could come with. 

I also didn't see issues in the issue tracker about it, but the fatal error is pretty evident if you run queries with the latest Laravel Lumen framework. 

Thank you for your time reviewing this pull request!

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -82,7 +82,7 @@ class QueryCollector extends PDOCollector
     {
         // If the query argument is an instance of QueryExecuted the rest of the
         // arguments are probably wrong types and are going to crash
-        // the colletor.
+        // the collector.
         if ( $query instanceof \Illuminate\Database\Events\QueryExecuted )
         {
             $bindings = $query->bindings;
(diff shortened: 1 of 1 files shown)


barryvdh:
THanks!

Naminator:
I'm sorry, can you revert the fix. I saw the greater issue. 
You have a weird conditional -> IF if (version_compare($this->version, '5.2.0', '>=')) with which you hope to catch the new version of laravel but for some reason my it is returning FALSE on my setup and I don't reapply the fix. 

I'll look into it and submit a new pull request once I figure what went wrong. Can please revert the pull request. I'm really sorry :(

barryvdh:
Reverted.

There is a `isLumen()` and `checkVersion()` helper if you want.

PR #469, opened 2016-02-16
Fixes #403 - Need to reflash the session.

### LINKED ISSUE #403: Loosing session flash data on ajax post request.
I have a form that is loaded by ajax, the controller giving back the form html is also saving some data as flash session data. If the form data is then send by ajax post request I can't retrieve the flashed data...

Disabling the debugbar plugin (dev-master) helps. Any idea?

bram1028:
This is a fix for issue #403.

This was lost when the package was updated to L5. The session needs to be flashed in our controller, otherwise the flash variables get lost once the package requests it's assets


barryvdh:
Thanks!

PR #487, opened 2016-04-11
Ported Session Flash bug fix from latest version

bgurney:
I needed to back port your Session Flash bug fix to a version of the package that runs on PHP 5.4.*

Feel free to merge if you think it'll be useful.

### DIFF
--- a/src/Controllers/BaseController.php
+++ b/src/Controllers/BaseController.php
@@ -2,6 +2,7 @@
 
 use Barryvdh\Debugbar\LaravelDebugbar;
 use Illuminate\Routing\Controller;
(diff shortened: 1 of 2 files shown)


barryvdh:
To where is this backported? You're targetting master, right?

bgurney:
Hey, we had to fork from v2.0.6 because this was the last version to support PHP >=5.4.0. I think some of the later versions state PHP >=5.4.0 in composer but they use the static class name resolution ::class syntax, which was introduced in 5.5.

barryvdh:
Is this still relevant?

bgurney:
Not for us as we've managed to move on now.

PR #490, opened 2016-04-20
add db query length limit option

devypt:


barryvdh:
Wouldn't it be better to update the layout to make it fit better, or to substr the parameters instead (eg. for 100 ids in a whereIn)

devypt:
I agree with substr the params suggestion , i will push commit for this.

aik099:
During query parameter substitution we can see if an array is coming in and display only first X entries and `...` after that. That would break query (if user decides to execute it by hand), but surely would save lots of space.

@devypt , any update?

barryvdh:
Or should we do something with css? To truncate long queries, but show the entire thing of hover?

PR #491, opened 2016-04-24
Delegate cache busting query string to the UrlGenerator

stanfieldr:
This is PR #379 except updated to work with Laravel 5 :)


barryvdh:
Does this work on all L5 versions?

stanfieldr:
Yep. Works on 5.1 and 5.2

PR #484, opened 2016-04-01
fix: Route caching

NelinD:
Route caching when the `artisan route:cache` command execute.

PR #495, opened 2016-05-02
ServiceProvider should not add routes if DebugBar is disabled

wiedem:
DebugBar currently adds its routes even if it is disabled, i.e. 'dead' routes are added.

Furthermore Laravel provides a method `runningUnitTests` to check if the app is running unit tests. This test should be used to skip the DebugBar activation during tests.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -21,12 +21,12 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
     {
         $configPath = __DIR__ . '/../config/debugbar.php';
         $this->mergeConfigFrom($configPath, 'debugbar');
-        
+
         $this->app->alias(
             'DebugBar\DataFormatter\DataFormatter',
             'DebugBar\DataFormatter\DataFormatterInterface'
         );
-        
+
         $this->app->singleton('debugbar', function ($app) {
                 $debugbar = new LaravelDebugbar($app);
 
@@ -39,7 +39,7 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
                 return $debugbar;
             }
         );
-        
+
         $this->app->alias('debugbar', 'Barryvdh\Debugbar\LaravelDebugbar');
 
         $this->app['command.debugbar.clear'] = $this->app->share(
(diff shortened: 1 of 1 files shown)


barryvdh:
This was done to fix route caching. What is the problem with dead routes?

wiedem:
> This was done to fix route caching.

I guess you're referring to https://github.com/barryvdh/laravel-debugbar/pull/484 ?
My PR doesn't really break the route caching functionality.

There's only one scenario where DebugBar doesn't work because of the route caching:
1. you either don't have DebugBar installed or DebugBar is disabled
2. `route:cache` is called
3. Debugbar is being installed or activated
The cached routes now don't have the routes needed by DebugBar, because you cannot dynamically add routes after they were cached - that's a well known limitation of Laravel. That's like modifying `routes.php` **after** a call to `php artisan route:cache` and wondering why you don't get the modified routes.
#484 doesn't really fix that either btw. If you have your routes cached before you install DebugBar and then install it, the DebugBar routes will always be missing unless you invalidate or rebuild the route cache.

That being said - no matter what you do, if you change routes (via `routes.php` or by adding a new service provider) and use route caching you **always** need to invalidate or rebuild the route cache. That's the responsibility of the developer and not of a service provider in Laravel.

My suggestion would be to rather add that to the documentation. Whenever you add/remove, enable or disable DebugBar the route cache has to be updated.

What you could do however is add a check in the `boot` method which calls `$this->app->routesAreCached()` and which checks if the needed routes are missing. You could then display a warning that the route cache needs to be rebuilt.

> What is the problem with dead routes?

I hope that's a rather rhetorical question. Production environments should never expose debug interfaces or contain any 'dead' interfaces of any kind including routes. That's crying for exploits and problems of all sorts.

barryvdh:
The problem with route caching was that the routes were never added when you were in the console, so whenever the cache command ran, it skipped the debugbar routes, so updating is impossible.

But indeed, it should not add the routes when disabled altogether.

barryvdh:
Sorry I misread about the caching thing. This does indeed not break that and does fix a potential future issue, introduced by #484  (Altough the debugbar status is checked before any data is returned in the routes, it's best to avoid this altogether)
Thanks!

wiedem:
> Sorry I misread about the caching thing. [...] Thanks!
No worries! You're welcome. Great work all around btw. DebugBar is a really great tool to debug Laravel apps.

May I still suggest you add a short section in the readme about the route caching? Some users might not be aware that they need to update their route cache after adding DebugBar or after enabling/disabling it.

barryvdh:
Sure, submit a PR if you want :)

PR #506, opened 2016-05-26
Setting 'enable' by .env file

GertjanRoke:
I like the DebugBar but when I put the product online (live), you need to manually turn off the DebugBar.
Maybe this is a nice way to do so.


GrahamCampbell:
This is a breaking change for anyone who's not published the config file.

barryvdh:
If you set it to null, it should listen to the app.debug value by default, see https://github.com/barryvdh/laravel-debugbar/blob/5056a3878db4ab7ac232d5a075d54aaa590c2d27/src/ServiceProvider.php#L69-L71

I've never had to change the config the enable/disable.

aik099:
Replaced by #629 (merged). This one can be safely closed.

PR #508, opened 2016-06-05
Bugfix for when sql param is not a questionmark

melkorr:
When debugbar.options.db.with_params = true, and when the sql being run is something like ``` SELECT * FROM users WHERE id = :id ```,  the :id param is not replaced.


barryvdh:
Thank!

Issue #497, opened 2016-05-11
PHP 7 Compatibility


iwex:
I have such exception: 
`FatalThrowableError in ExceptionsCollector.php line 28:
Type error: Argument 1 passed to DebugBar\DataCollector\ExceptionsCollector::addException() must be an instance of Exception, instance of ParseError given, called in /home/iwex/vagrant/chicago/vendor/maximebf/debugbar/src/DebugBar/DataCollector/ExceptionsCollector.php on line 32`

barryvdh:
We can't typehint the new Errors right? So no typehinting at all then?

iwex:
Hmm, is it possible to create new branch for php7. For typehinting you can use Throwable interface.

marktopper:
@iwex did you create an issue about this on [`maximebf/php-debugbar`](https://github.com/maximebf/php-debugbar)?

iwex:
@marktopper no :)

barryvdh:
So perhaps just like this; remove the typehint and Check at runtime.
https://github.com/Seldaek/monolog/commit/481fc4cfa17808581d113673e7dd002b06dcc100

tje3d:
Works fine with: PHP 7.0.4

vjoao:
@tje3d 
It's not working with PHP 7.0.5-2... should I downgrade?

tje3d:
@vjoao 
hmm, didnt tried that, im not sure with 7.0.5

marktopper:
It should work in PHP 7.x when exceptions are thrown, but not when errors are thrown.

vjoao:
It was a mistake on my side. Exceptions are namespaced and I forgot to use them.

alejandrozorita:
Hi!
I have the same fail with PHP 7.0.7...

ThomasRalee:
@alejandrozorita 
Same am facing problems with PHP 7.0.8

barryvdh:
Can you try disabling collectors until it doesn't happen, so we can narrow down which one it is?
Or if it's the debugbar itself (without collectors)

Sly-Ripper:
Is there a reason this isn't fixed yet? Still the same error as OP, just line 28.

siegerhansma:
@barryvdh Just tried this. It seems to be the Exceptions collector that is giving issues.

DrudgeRajen:
Same problem with 7.0.9

iget-master:
Same problem here... 

[...]

PR #510, opened 2016-06-10
Added check to ensure exceptions passed to debugbar are compatible with type-hint in php7 #497

jamiehd:


jamiehd:
Sorry, forgot the phpdoc hint in the other one.

barryvdh:
But this is still  listed to 2 levels deep.

jamiehd:
Nope the function is recursive.

barryvdh:
But looking at https://github.com/symfony/debug/blob/master/Exception/FatalThrowableError.php it will remove the previous exceptions right?
So should we just check if the actual exception is Throwable, and if so convert that.
And in php-debugbar check the previous exceptions, when added? So that we keep al information?

jamiehd:
Yeah I've actually left a working implementation of that in the other pull request. I thought this was a better solution as it didn't make the symphony debug package a dependency of php-debugbar and also didn't require changes to both packages. I'm happy to submit my other fix as a pull request to each package if you'd rather. I've detailed it here: https://github.com/maximebf/php-debugbar/pull/276#issuecomment-225018643

barryvdh:
I think that's probably better. But add a check for instanceof Throwable, instead of just `else`

jamiehd:
Ok. And if it's not throwable do nothing with it?

barryvdh:
I guess, but that *shouldn't* happen. It's just a bit more readable/clear what the intention is.

barryvdh:
Or change it around like this:

```php
if($previous instanceof Throwable) {
	$previous = new FatalThrowableError($previous);
}
$this->addException($previous);
```

jamiehd:
I thought about that, but do older versions of php have Throwable?

barryvdh:
Oh sorry, that doesn't make sense, throwable matches all: https://3v4l.org/9n7DB

But we can check for \Error, even on php5: https://3v4l.org/jrKjc

jamiehd:
Ok that makes sense to me, thanks.

PR #517, opened 2016-06-16
adjust middleware order

sodoojin:
in laravel 5.1, debugbar makes previous url to asset url like http://example.com/_debugbar/assets/javascript?1456726057

debugbar assets are handled by controller. 
so, \Illuminate\Session\Middleware\StartSession is called and then _previous variable in session is changed by storeCurrentUrl function.

that's why "return back();" redirects to debugbar asset page.

so, i moved debugbar middleware to top level of middlewares.
after that it works well for me.

if it still has problem. please let me know. i will fix it again.

thank you.

laravel 5.2 is fine. but laravel 5.1 is not
i didn't check the all version but i think there is a same bug under 5.1 version.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -151,7 +151,7 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
     protected function registerMiddleware($middleware)
     {
         $kernel = $this->app['Illuminate\Contracts\Http\Kernel'];
-        $kernel->pushMiddleware($middleware);
+        $kernel->prependMiddleware($middleware);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sodoojin:
i used to set enable flag by using closure.
but it doesn't work after update debugbar.

you dropped it? or it is a bug?

barryvdh:
I think I dropped it because of the config caching, which doesn't support closures.

aik099:
Maybe this would be solved by #569, since DebugBar data would be loaded via AJAX and it doesn't change `_previous` session attribute.

barryvdh:
Is this still relevant?

PR #518, opened 2016-06-25
Simplify array syntax

Tepster:


barryvdh:
Thank!

barryvdh:
Hmm. Did you do this by hand?
Errors in #520 and #521 
Not sure if there are more now, perhaps better to revert otherwise :/

Tepster:
Sorry. I've run syntax checker and there are no more such errors.

PR #524, opened 2016-07-18
Allow callback to determine `enabled` config value

### LINKED ISSUE #523: Debug false but true for some IP's
Is it possible to extend Debugbars configuration to create a possibility to hide the Debugbar for everyone except an array of IP's?

Sometime I have to debug on a new (live) server and I don't want anyone to see the debug bar. I know I can use the log, but I rather see the styled version of my error's and warnings... :-)

robinvalk:
When callbacks are allowed, you can for example enable or disable the debugbar for specific ips.

Resolves #523

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -64,7 +64,7 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
         $this->publishes([$configPath => $this->getConfigPath()], 'config');
 
         // If enabled is null, set from the app.debug value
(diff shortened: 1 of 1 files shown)


robinvalk:
Closures aren't supported, https://github.com/laravel/framework/issues/9625
Needs a different solution then...

barryvdh:
You can conditionally add the serviceprovider in your own provider and set `debugbar.enabled` to `true`.
Or always register it, but conditionally call `$debugbar->enable()`

PR #534, opened 2016-08-02
option to only collect slow queries

bdanchilla:
check if a new config setting `debugbar.options.db.slow_threshold` is set. if so, only add the current query to the collector if it is slower then this setting

ie)

```
 'options' => array(
        'db' => array(
            'slow_threshold'    => 100, //milliseconds a query must be slower then to be collected
        ),
```


barryvdh:
I guess this could work, but it would need an entry in the config file also.

barryvdh:
I do think this is interesting, but the time is not most important in the development area, so not sure how relevant it is in practice.

PR #542, opened 2016-08-31
Fix for displaying middleware names

memu:
If you assign multiple middlewares to a route, the DebugBar shows the indexes of middlewares, not names.

No need for `getMiddleware()` method to check if route has multiple middlewares, since `\Illuminate\Routing\Route@getAction()` method does it for us.

route

### DIFF
--- a/src/DataCollector/IlluminateRouteCollector.php
+++ b/src/DataCollector/IlluminateRouteCollector.php
@@ -74,26 +74,7 @@ class IlluminateRouteCollector extends DataCollector implements Renderable
             $result['file'] = $filename . ':' . $reflector->getStartLine() . '-' . $reflector->getEndLine();
         }
 
-		if ($middleware = $this->getMiddleware($route)) {
-		    $result['middleware'] = $middleware;
-		}
-
-
-
         return $result;
-    }
-
-    /**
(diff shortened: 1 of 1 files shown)


barryvdh:
Since when does it do that?

memu:
Line 52: `$action = $route->getAction();`

This method returns an array and it contains the middleware(s). Like:

```php
dump($action)
```

```php
array:7 [
	"middleware" => array:2 [
		0 => "web"
		1 => "setLocale"
	]
	"as" => "author.login"
	"uses" => "App\Http\Controllers\Author\Auth\LoginController@showLoginForm"
	"controller" => "App\Http\Controllers\Author\Auth\LoginController@showLoginForm"
	"namespace" => "App\Http\Controllers\Author\Auth"
	"prefix" => "/author"
	"where" => [ ]
]

```

barryvdh:
Is this still required?

RamonSmit:
Would be nice to have :-)

memu:
I use Laravel 5.4 & Laravel Debugbar 2.4.1 and it looks like everything is ok. Middleware names are displayed instead of indexes. But I have no idea about other versions.

Issue #539, opened 2016-08-25
problem with RTL languages


Palvaneh:
please add `direction: ltr;` to `div .phpdebugbar` in **laravel-debugbar.css**
thanks

barryvdh:
Can you submit a PR to https://github.com/maximebf/php-debugbar

PR #543, opened 2016-09-05
fixing display in RTL languages

boynet:
see https://github.com/barryvdh/laravel-debugbar/issues/539


barryvdh:
Does this fix it? https://github.com/maximebf/php-debugbar/pull/267

boynet:
yes same thing just targeting the root element

boynet:
Hi I think it's got forgotten :) would love to merge this simple fix.

there is really nothing to worry about as it's just reset to default values.
if you don't reset it and you have direction:rtl on the body than its got messed up 


see the docs for the default values:
[direction](https://developer.mozilla.org/en-US/docs/Web/CSS/direction)
[text-align](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align)

PR #547, opened 2016-09-08
Add optional request header to disable debugbar response injection

StevenBock:
This adds an optional header that you can add to your request to disable the debugbar data collection/response injection.  I looked through the code and could not find anything obvious to do this.

I experienced the initial issue with running the [Node Laravel Echo Server](https://github.com/tlaverdure/laravel-echo-server).  When joining a private channel, it sends an HTTP request to the broadcasting auth endpoints, which in turn replies "true" if the user can join the channel.  However, in a development environment with response injection enabled, the debugbar injection happens and the echo server is unable to parse the reply (it's expecting JSON) correctly.  Having a way to disable the injection with this request would remediate this issue.

I hope this is useful and if something is obviously wrong, any pointers would be great.  This is my #1 pull request on an open source project!

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -157,6 +157,19 @@ return [
 
     /*
      |--------------------------------------------------------------------------
+     | No Debugbar Injection Header
+     |--------------------------------------------------------------------------
+     |
+     | Specify an optional header you can add to individual requests to disable
+     | debugbar data collection and response injection.  This is useful when you
+     | have a certain request that must be unaltered.
+     |
+     */
+
+    'no_injection_header' => 'X-NO-DEBUGBAR',
+
+    /*
+     |--------------------------------------------------------------------------
      | DebugBar route prefix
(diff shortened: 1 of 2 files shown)


barryvdh:
It shouldn't inject in Ajax requests or json respones, so are you sure your using the correct responses?

StevenBock:
Hmm, it looks like it's sending a standard POST request.  [https://github.com/tlaverdure/laravel-echo-server/blob/master/src/echo-server.ts#L451](Code in question)

Might be we can send it as an AJAX request and skip the need for this.  I'll do some investigation.

StevenBock:
Ok, so setting `options.headers['X-Requested-With'] = 'XMLHttpRequest';` does indeed fix the issue with the response being modified.  So I suppose this pull request is not needed, although it might be nice to have some way to completely disable the debugbar without having to send your request as an AJAX call.

Thanks for your time!

PR #557, opened 2016-10-25
css/js loaded at any protocol case

ysugimoto:
Debugbar doesn't working in under the HTTPS-HTTP reverse-proxy.

For example,

```
HTTPS (AWS ELB) ->HTTP (Laravel)
```

In this case,  browser blocked debugbar's css and js on different protocol between https and http.
So, remove protocol prefix at loading css and js, it will success to load these assets.

### DIFF
--- a/src/JavascriptRenderer.php
+++ b/src/JavascriptRenderer.php
@@ -43,7 +43,10 @@ class JavascriptRenderer extends BaseJavascriptRenderer
 
         $jsRoute = route('debugbar.assets.js', [
             'v' => $this->getModifiedTime('js')
-       ]);
+        ]);
+
+        $cssRoute = preg_replace('/\Ahttps?:/', '', $cssRoute);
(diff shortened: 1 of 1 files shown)


barryvdh:
You should set up your trusted proxies correctly probably, so Laravel will generate proper URLs with the correct scheme.

ysugimoto:
@barryvdh Yes, Laravel will generate proper URLs. But In my environment, HTTPS connection handles AWS ELB, and transport to EC2 working Laravel (with Apache HTTP) on inside networks. 

At this time, Laravel generates `http://` URL on `route()` function, perhaps because of `Request::isSecure()` returns false. for that reason, browser connecting with https url, but Laravel generated url is http.

In any case, css and js loading path is starting with `//`, no write ptotocols, I think that there is no need to worry about protocols. What do you think about it?

barryvdh:
I think I'd rather rely on Laravel behavior, but I get the point.
Did you try https://github.com/fideloper/TrustedProxy to fix all your route() calls?

Issue #566, opened 2016-11-22
Add xdebug links to views / controllers


amenk:
It would be nice to be able to click the file references

PR #567, opened 2016-11-22
Add Xdebug Link functionality

amenk:
https://github.com/barryvdh/laravel-debugbar/issues/566

### DIFF
--- a/src/DataCollector/ViewCollector.php
+++ b/src/DataCollector/ViewCollector.php
@@ -80,12 +80,18 @@ class ViewCollector extends TwigCollector
             $params = $data;
         }
 
-        $this->templates[] = [
+        $template = [
             'name' => $path ? sprintf('%s (%s)', $name, $path) : $name,
             'param_count' => count($params),
             'params' => $params,
             'type' => $type,
         ];
+
+        if ( $this->getXdebugLink($path)) {
+            $template['xdebug_link'] = $this->getXdebugLink($path);
+        }
+
+        $this->templates[] = $template;
     }
 
     public function collect()


amenk:
see https://github.com/maximebf/php-debugbar/pull/303

aik099:
Would be great feature, once merged. Right now in traces (see image below) I see exact name of file and line where to go, but clicking on it does nothing. Would be great if it would open that file in PhpStorm automatically.



Also we should think of path mapping, because I have Laravel running inside VM where paths are `/home/username/web/project-name`, but PhpStorm is running locally, where paths are `/Volumes/web/project-name`.

amenk:
Path mapping should usually be handled by the IDE. PhpStorm has a configuration for that.

aik099:
It has mapping, when debugging connection comes in. But do they work, when you have link like `phpstorm://....` with path on VM?

amenk:
The mapping is configuration wise not bound to debugging. It says in general "mapping" ... so it should :) I do not see this mapping logic to be a part here in the PHP project.

aik099:
OK. Then I guess we need to wait and see this merged.

decadence:
phpstorm:// works only on Mac.
For cross platform: `"http://localhost:63342/api/file/{$file}:{$line}"`

aik099:
Just checked and path mapping has no effect. Reported issue for PhpStorm: https://youtrack.jetbrains.com/issue/WI-37690

aik099:
The JetBrains team (PhpStorm developers) confirmed, that path mapping is only for debugging, not for remote links.

PR #569, opened 2016-11-25
Option to use debugbar with content-security-policy headers

Ricky-rick:
The debug bar wasn’t compatible with a content security policy that doesn't allow inline-script.
I made it possible to load the init script through a route and made use of the option to load the data with an ajax request. 

You need to add "img-src:'self' data:" to your content security policy header and storage needs to be enabled.

The lines for ‘jQuery.noConflict(true); ’ are moved to the getJsInitializationCode because those needs to be at the scripts file. All other changes are new lines.

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -591,7 +591,7 @@ class LaravelDebugbar extends DebugBar
             } catch (\Exception $e) {
                 $app['log']->error('Debugbar exception: ' . $e->getMessage());
             }
-        } elseif ($app['config']->get('debugbar.inject', true) && $app['config']->get('debugbar.csp-compatible', true)) {
+        } elseif ($app['config']->get('debugbar.inject', true) && $app['config']->get('debugbar.csp-compatible', false)) {
             try {
                 $this->injectDebugbarCSP($response);
             } catch (\Exception $e) {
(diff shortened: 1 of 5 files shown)


src/LaravelDebugbar.php
```diff
@@ -591,6 +591,12 @@ public function modifyResponse(Request $request, Response $response)
             } catch (\Exception $e) {
                 $app['log']->error('Debugbar exception: ' . $e->getMessage());
             }
+        } elseif ($app['config']->get('debugbar.inject', true) && $app['config']->get('debugbar.csp-compatible', true)) {
```
Ricky-rick: You are right. I overlooked this default value. I will change it next Monday and open new pull request.

marktopper: You know you can just make a new commit to your branch and then it will change this pull request, right?

Ricky-rick:
Any info why this pr isn't merged?

marktopper:
Ping @barryvdh

aik099:
Few changes:

1. this needs to be rebased
2. it seems, that it should target 2.4 branch (default branch in repo), rather than `master`

andersLAL:
Any hope for this one? I would love to be able to abandon my own fork...

barryvdh:
So this will (when enabled) always load the script with ajax? Could be nice for performance even.

PR #571, opened 2016-12-05
env

jrean:
Please :-)


barryvdh:
Currently it's the same as APP_DEBUG, is that not ok for you?

jrean:
Your package and your work are amazing! Adding the `env` just allow all the possibilities :-)

aik099:
Replaced by #629 (merged already). So this one can be safely closed.

PR #577, opened 2016-12-17
Update Installation

iGusev:
Added `--dev` param


barryvdh:
Then the rest should also be updated, otherwise it's breaking in production

iGusev:
but its dev util and unnecessary dependence in production, isnt it?

barryvdh:
Yes

PR #578, opened 2016-12-19
Add Left to right direction to css

MeeDNite:
So it doesn't mess up styles on RTL pages.


MeeDNite:
Oh, There was another PR already.
My bad.

PR #583, opened 2017-01-06
Update for 5.4

martindilling:
There might be more already, and there might come more in the future.
But this seems to at least fix the breaking changes that prevents the application to boot when you install the DebugBar in a new Laravel 5.4 installation.

Commits with breaking changes that affected this package: (that I know of)
Log: https://github.com/laravel/framework/commit/57c82d095c356a0fe0f9381536afec768cdcc072?diff=split
Session: https://github.com/laravel/framework/commit/66976ba3f559ee6ede4cc865ea995996cd42ee1b?diff=split

PR #584, opened 2017-01-07
[5.4] The HttpDriver gets a SessionManager instead of Session\Store, so instanceof check didn't work

martindilling:
Fix for the error mentioned #583.
I can't really figure out why it worked for me before, looking at this now.
But the SymfonyHttpDriver is receiving a SessionManager and not directly the object implementing the new Session interface.
Not sure if this is the best or will work for all cases, but it's a start.


oeholmen:
The session is not updated in Laravel 5.3 after this update. All requests are stored and will load when they are clicked in the debugbar panel.

debugbar

PR #606, opened 2017-02-16
Support phpredis php extension

Except10n:


barryvdh:
Is this backward compatible with older versions?

Except10n:
yes, [framework/src/Illuminate/Redis/Connections/Connection.php](https://github.com/laravel/framework/blob/5.4/src/Illuminate/Redis/Connections/Connection.php) added after 5.4

barryvdh:
But it still needs to work pre-5.4

Except10n:
```
if (is_a($client, 'Illuminate\Redis\Connections\Connection', false)) {
    $client = $client->client();
}
```
It works fine in pre-5.4
But [php-debugbar](https://github.com/maximebf/php-debugbar/pull/319) still need to be modified, it only support predis client or interface

Issue #483, opened 2016-04-01
Track where database queries are coming from


rjosephporter:
Is there a way to track what class/file a database query was called? 

This will surely help a lot.

barryvdh:
Did you try the db backtrace option in the config?

rjosephporter:
Thanks! Exactly what I needed :+1:

sebdesign:
May I raise a similar issue here?

I find the backtrace option very very useful, but I often need to backtrace a few more steps down the stack. E.g. I'm dealing with a legacy codebase, and I have several queries that are executed on my eloquent models (on relations or other methods).

That usually is a code smell that indicates that eager loading should be applied **before** calling those methods. But due to the *spagghetti* nature of that codebase, it's not easy to find the controllers or the intermediate classes which call those methods on the models. If I could backtrace a few levels back that would be very helpful.

Specifically, the `QueryCollector@findSource` could collect all the meaningful stack lines (instead of returning one the first one). The `debug_backtrace` method accepts a `$limit` parameter, which could be set in the config.

Then on the queries tab, the first backtrace result would be displayed (like it is currently). but when expading on a query, the stack trace could be listed along with the SQL and the bindings too.

barryvdh:
Hmm could be interesting. Just a bit worried about the information overload..

sebdesign:
I understand your concern, but as a designer (also) I could give it a try to make the information look neat.
Currently, the way the query information is displayed on expansion is a bit clunky. The params `<table>` is prepended to the left and is causing the hints to look weird.

barryvdh:
Ok, I'm open for suggestions and improvements :)

kirkbushell:
Was there any movement on this? 5 lines of stack trace is nothing when debugging queries :)

PR #608, opened 2017-02-18
Improve the queries tab

sebdesign:
laravel-debugbar-queries

- Collect database transactions events (begin, commit, rollback).
- Display the stack of the backtrace when a query item is expanded.
- Model binding is detected explicitly in the `Router#substituteBindings` method.
- The backtrace now detects middleware.
- Redesigned the params table to display bindings, hints and backtrace nicely.
- The data formatting methods have been extracted to a `QueryFormatter`.

See https://github.com/barryvdh/laravel-debugbar/issues/483

P.S.: In the screenshot I have replaced the fonts with a system font stack. I plan to submit a PR to the [php-debugbar](https://github.com/maximebf/php-debugbar) repo about that.

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -124,6 +124,7 @@ class QueryCollector extends PDOCollector
 
         $this->queries[] = [
             'query' => $query,
+            'type' => 'query',
             'bindings' => $this->getDataFormatter()->escapeBindings($bindings),
             'time' => $time,
             'source' => $source,
@@ -208,7 +209,6 @@ class QueryCollector extends PDOCollector
             'index' => $index,
             'namespace' => null,
             'name' => null,
-            'file' => null,
             'line' => isset($trace['line']) ? $trace['line'] : '?',
         ];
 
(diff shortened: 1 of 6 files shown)


sebdesign:
I just noticed that the transaction event listeners are missing from this commit. I will add them in the next commit!

barryvdh:
Looks nice, one thing I'm not sure about is the pre rendering of HTML in the results, instead of modifying the query widget in php-debugbar

sebdesign:
I agree with you on that point.

I tried to extend the sql widget, so I could manipulate the params table, but couldn't get it to work.
Something about a constructor not found, or the SQLWidget being undefined. Perhaps you help me through that?

sebdesign:
It seems I can't simply extend `SQLQueriesWidget`, so I have to extend the base `Widget`. I'll work on that and will update the PR.

barryvdh:
Let's give this a go :)

barryvdh:
Looks good to me. I've enabled the backtrace by default now (including some other collectors).

I keep seeing a lot of public/index.php backtraces, so gonne try to filter those out as well.

barryvdh:
Oh was 3rd party repo; fixed with https://github.com/barryvdh/laravel-debugbar/commit/e1ee861ebaab6537dbe295a13deba55f6bb677ec

Thanks!

sebdesign:
Thanks for merging this!

While implementing this, I was thinking maybe a `Debugbar::backtrace()` method would be handy to display the given stack in a similar fashion on the messages tab. I can prepare a PR if you want.

barryvdh:
Hmm, not really sure. Wouldn't you usually know where it comes from?

sebdesign:
Usually, yes. But when inspecting a legacy/unfamiliar codebase, I might want to trace down a function/method call. Using `debug_backtrace` or `debug_print_backtrace` would render a stack from index.php until my function is invoked, so I can figure out the intermediate calls. It would be nice to print that stack in the debugbar instead of the window.

PR #623, opened 2017-04-02
Add route domain config option

pinepain:
This PR adds `debugbar.route_domain` config option to allow specifying a custom domain name where debugbar is served from.

The use case is to deal with wildcard route conflicts.

This PR is fully BC compatible.

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -180,4 +180,13 @@ return [
      */
     'route_prefix' => '_debugbar',
 
+    /*
+     |--------------------------------------------------------------------------
+     | DebugBar route domain
+     |--------------------------------------------------------------------------
+     |
+     | By default DebugBar route served from the same domain that request served.
+     | To override default domain, specify it as a non-empty value.
+     */
+    'route_domain' => null,
 ];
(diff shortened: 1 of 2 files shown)


pinepain:
ping @barryvdh

pinepain:
@barryvdh Is there something I can do in advance to help this changes appears in master?

barryvdh:
Aren't wildcard conflicts resolved when you put the debugbar last?

barryvdh:
Oh wait, other way around, accorrding to readme:

https://github.com/barryvdh/laravel-debugbar#installation
>If you use a catch-all/fallback route, make sure you load the Debugbar ServiceProvider before your own App ServiceProviders.

pinepain:
Hi! Thanks for quick answer and for suggestion. For sure, it's a working solution at this time which I'm using for now. The problem I'm trying to solve is to make 100% sure that main application routes won't be ever overlap with development routes. I deal with multi-tenant application where end-user may define any route, so using separate (sub)domain is a sort of solution to make sure no overlap happens. While it's quite rare use case, I think having domain option may do the job for someone like me and it should cause no harm as it preserves BC with current implementation. What do you think?

pinepain:
@barryvdh any chances to merge this change?  Apparently, we could provide a way to customize whole route config array by adding some `route_config` array-type config item and then merge it with route config that calculated within service provider?

pinepain:
Thanks!

PR #633, opened 2017-05-15
Validate $recaller in MultiAuthCollector.php

### LINKED ISSUE #249: Undefined Offset 0
I'm using laravel 4.2 and I was trying to display a variable that I'm getting from my Cache. As you can see my variable is an object that I want to see in the DebugBar :
$var = Cache::get('data');
$var = json_decode($var);
Debugbar::info($var);

The error this gives is : 
Undefined offset: 0 (in line 73 : vendor/symfony/var-dumper/Symfony/Component/VarDumper/Cloner/VarCloner.php)

but the exception was raised from :  Debugbar::info($var);

Do you have any idea where this is coming from ? knowing that it was working properly before, and I haven't made any updates to the packages

n0f:
fixed #249

### DIFF
--- a/src/DataCollector/MultiAuthCollector.php
+++ b/src/DataCollector/MultiAuthCollector.php
@@ -1,8 +1,9 @@
 <?php
 
 namespace Barryvdh\Debugbar\DataCollector;
-use Illuminate\Contracts\Auth\Guard;
+use Illuminate\Auth\Recaller;
 use Illuminate\Auth\SessionGuard;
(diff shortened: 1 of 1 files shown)


lucasgameiro:
After this commit, the support need to be updated to laravel 5.4+

veneliniliev:
this broke all sites with laravel 5.3!

barryvdh:
Can someone submit a PR to check the version/class?

PR #638, opened 2017-05-24
Validate $recaller in MultiAuthCollector.php

n0f:
Fixed #633 .
Validate $recaller in MultiAuthCollector.php without the Recaller class.
I modified it to work the same as when using the Recaller class.

### DIFF
--- a/src/DataCollector/MultiAuthCollector.php
+++ b/src/DataCollector/MultiAuthCollector.php
@@ -1,9 +1,10 @@
 <?php
 
 namespace Barryvdh\Debugbar\DataCollector;
-use Illuminate\Auth\Recaller;
+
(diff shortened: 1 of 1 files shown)


barryvdh:
This works on both 5. 3 and 5.4? Did you try both?

n0f:
I do not have a test case for valid integration testing.

This series of modifications fixes a bug in the `resolveUser()` method of the MultiAuthCollector class.
In the last change, I made a mistake and used the Recaller class that depends on Laravel 5.4+.
In this change, we have resolved this method's dependency on the Recaller class.

Since neither change affects the specification of the MultiAuthCollector class, we only executed unit tests.
This method now depends only on `Str::contains` (Laravel 4+), so there is no longer any effect from the difference in version.

PR #641, opened 2017-05-31
Open Handler Loaded at any Protocol Case

adamyi:
Similar to #557 , currently if laravel is running in HTTP protocol behind a HTTPS reserve proxy, it will generate route of open handlers as HTTP. However, many browsers disallow this mixed content for security concerns. So, remove protocol prefix at the open handler, it will manage to load those to deal with AJAX requests.


barryvdh:
Should be fixed with correctly setting the trusted headers

aik099:
Actually you need to replace `http://` and `https://` with just `//`.

I wonder if there any way to build URL to the route without protocol at all from Laravel?

PR #643, opened 2017-06-01
Support Auto-Discovery

taylorotwell:
This will make the package work with auto-discovery in Laravel 5.5. Enjoy :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,14 @@
     "extra": {
         "branch-alias": {
             "dev-master": "2.3-dev"
+        },
+        "laravel": {
+            "providers": [
+                "Barryvdh\\Debugbar\\ServiceProvider"
+            ],
+            "aliases": {
+                "Debugbar": "Barryvdh\\Debugbar\\Facade"
+            }
         }
     }
 }


Francismori7:
No way that's awesome!!

barryvdh:
Thanks!

barryvdh:
@taylorotwell I tagged this as `v2.4.0`. So if you want you could updated your blog post(https://medium.com/@taylorotwell/package-auto-discovery-in-laravel-5-5-ea9e3ab20518)
to use remove `dev-master`, or perhaps better even: `composer require barryvdh/laravel-debugbar --dev`, to install it in dev only :)

lukepolo:
Although early, will there be a way to disable this autoloading?

```
Use case :  
I only want to add the service provider / facade in development only 
```

guess we may have to wait till 5.5 is released to find out that answer

barryvdh:
See my comment above, when you require with - - dev it is only installed and loaded on dev.

pk-informatics:
@lukepolo you can also disable autoloading - see the highlighted part in taylors blogpost about autoloading: https://medium.com/@taylorotwell/package-auto-discovery-in-laravel-5-5-ea9e3ab20518

lukepolo:
Yup thank you , missed those parts shouldn't have skimmed it

mikimaine:
What if I want to disable it only for testing some things without removing it completely?  Currently, I can go to app.php and comment out the service provider.

barryvdh:
Debugbar only loads when debug = true

mikimaine:
Ya for debug bar package there is that ENV value. What I want to know is for other packages, How will the auto discovery handles that.

barryvdh:
You can disable it, see the blog

marktopper:
@mikimaine 
image

cheerego:
Such nice feature 😀😀😀

adampatterson:
Nice, should also help with updates/upgrades when your plopping configs between projects.

nelson6e65:
Hmmm... it's nice is loaded only in local development, but.. what about aliases usage?

For example, in my code I can use the `Debugbar` facade alias, for its methods. So, in production **it will not trigger a non-existent class error?**

I'm still have to check if I'm on local or production in order to use it. So, it needs to be available on production, even if it is disabled. u.u

barryvdh:
You shouldnt leave those calls in your production code, no.

PR #657, opened 2017-06-14
Change case-name of class (allows composer optimizing)

### LINKED ISSUE #654: Class name improperly cased in 1.8 branch
Debugging an old L4.2 project and I installed this package.  I also use `composer dump-autoload -a -o`, and get the error:

```
Class 'Barryvdh\Debugbar\LaravelDebugBar' not found
```

I think this has to do with the fact that the file is called `LaravelDebugBar.php` but the class is defined as

```php
class LaravelDebugbar extends DebugBar
                  ^
```
The `b` in `bar` should be capitalized, so the class name matches the file name's case.

Does that make sense?

cviebrock:
Fixes #654

### DIFF
--- a/src/LaravelDebugBar.php
+++ b/src/LaravelDebugBar.php
@@ -42,7 +42,7 @@ use Symfony\Component\HttpFoundation\Response;
  * @method void debug($message)
  * @method void log($message)
  */
-class LaravelDebugbar extends DebugBar
+class LaravelDebugBar extends DebugBar
 {
     /**
      * The Laravel application instance.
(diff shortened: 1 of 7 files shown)


barryvdh:
I would rather do this the other way around, same as in master. https://github.com/barryvdh/laravel-debugbar/blob/master/src/LaravelDebugbar.php

cviebrock:
Renamed!

cviebrock:
Thanks @barryvdh !  You'll tag a new release at some point?

Issue #634, opened 2017-05-15
[Proposal] Cache Tab


decadence:
Hello. It's not an issue but proposal.
It would be great to have Cache Tab where we can see Cache values / keys which has been involved in Request.
You can use events from Illuminate\Cache\Events namespace. There are events for get, forget, set, missed value.

decadence:
Isn't anybody interested in this feature?

decadence:
@barryvdh any thoughts on this?

decadence:
Merged

PR #678, opened 2017-07-21
add cache tab

adelf:
cache-tab

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -116,6 +116,7 @@ return [
         'logs'            => false, // Add the latest log messages
         'files'           => false, // Show the included files
         'config'          => false, // Display config settings
+        'cache'           => false, // Display cache events
     ],
 
     /*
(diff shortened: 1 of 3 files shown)


decadence:
https://github.com/barryvdh/laravel-debugbar/issues/634

barryvdh:
Why are you serializing the values?

adelf:
mm.. I did it week ago. I didn't like something when there was pure values. I'll check it on Monday.

adelf:
I tried without serialize and get '[object Object]' for each class which stored in cache. Thats not very informative. I wanted to see what exactly goes to cache. But I used laravel-debug-bar 2.2. Maybe in current version it's much better? (as I see there are VarDumpers instead of ValueExporters)
Anyway, @barryvdh it's up to you. I can remove serializing, or make it optional...

barryvdh:
It should use the vardumper indeed, similar to the message or view objects.

adelf:
Yes. I misunderstood something. Fixed. Thanks.

barryvdh:
Thanks

decadence:
Thanks

barryvdh:
I pulled this in branch 2.4

adelf:
Yes. It should support all 5.* versions.

PR #680, opened 2017-07-25
fix cache tab, support laravel>=5.4 event system

adelf:
they changed listener params for wildcard events in 5.4...


barryvdh:
The `master` branch is 5.5 only, so doesn't need this checks, does it?

PR #698, opened 2017-08-17
add load_debugbar flag and update the doc

wangxiaomo:
sometimes i need to control debugbar at run time, i've tried following [the doc](https://github.com/barryvdh/laravel-debugbar#enablingdisabling-on-run-time),which means i need to do the following

```
# CANNOT disable debugbar in configuration, and make sure debugbar enabled
# disable debugbar at app/Http/Controllers/Controller.php
# and then enable debugbar somewhere at run time
```

that's quite strange i think. so i made changes in this pr, hope you guys will like it.

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -7,11 +7,17 @@ return [
      | Debugbar Settings
      |--------------------------------------------------------------------------
      |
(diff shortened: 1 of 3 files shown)


barryvdh:
Don't think this should be required.

Issue #693, opened 2017-08-16
hide some secret & password ?


Grummfy:
Hello,
in \Barryvdh\Debugbar\DataCollector\SymfonyRequestCollector::collect we see that we have a loop that remove _KEY, _PASSWORD, _PW & _SECRET and hide them.
I think it could be interesting to have a configuration to add extra fields to hide.

For example, on clever cloud, you have a "MYSQL_ADDON_URI" that is a dsn pdo url ready to use, so with login & password ;)

I imagine something like this in the configuration:
```php
// ...
'options' => [
  'request' => [
    'hide' => ['MYSQL_ADDON_URI'], // will  be removed from the collection
    'secrets' => ['SOME_ENV_VARS'], // will be replace by '******'
  ],
// ...
```

thanks

PR #699, opened 2017-08-19
Add custom request filter (#693)

profbiss:
Pull request to close issue #693

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -74,7 +74,7 @@ return [
      |
      */
     'error_handler' => false,
-    
+
     /*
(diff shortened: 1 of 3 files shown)


barryvdh:
Do we need both variants? Just the secret seems enough?

benyanke:
Any updates on when this might get pulled in?

For the record, I see a use for hide/vs/secrets. One hides things other devs wouldn't care about, and the other would be for things that matter, but just shouldn't have their content displayed.

barryvdh:
I think I'm gonna solve this differently but still like the idea.

PR #711, opened 2017-08-31
Ignore explain query for clockwork

yusukezzz:
Because the data structure is different from the sql type query.
Explain query has not `duration` and `connection` column, so raised `Undefined index` Notice.
This is only workaround.


aik099:
This happens for `EXPLAIN` statements that you trigger manually or `EXPLAIN` statements (if any) triggered by debugbar itself?

Issue #708, opened 2017-08-29
L5.4 - Does the doco need to be updated to conditionally load the service provider?


ashgibson:
The installation section of the docs recommend installing as a dev only dependency as per below:

composer require barryvdh/laravel-debugbar --dev

but then add service provider to config/app.php and debug bar will only be shown when debug is on.

Won't this cause an issue on production where dev dependencies are not installed, therefore the service provider will fail to load?

aik099:
You're correct. I guess the change was made to accompany Laravel 5.5+ service provider auto-discovery where you don't need to register service manually.

In any case the `--dev` switch must be removed, because:

1. provider should be added in all cases
2. the debug panel is only activated in debug mode, so no harm done

barryvdh:
The change is made on the 5.5 version. You don't need to register it manually so no need for conditionals. Also, it is very recommended to use it for development only.

aik099:
But what if you need to see something happening only on production servers and no way to replicate it locally?

barryvdh:
Then you install it without - - dev, but I wouldn't recommend that, at least not by default. I made the change because in 5.5 the package loads by default when installed with composer, so could be unexpected for people who are using it conditionally.

aik099:
I understand that of course. Since in Laravel 5.5 there is no way to avoid Laravel DebugBar service provider from being added automatically, then only way to hide it would be:

* either making it a dev dependency
* or disable it via config

If solution like I proposed in https://github.com/laravel/internals/issues/766 would be implemented, then having IP-based debug mode would allow to have debug bar always connected but enabled only for developers.

aik099:
From README.md:

> If you don't use auto-discovery, add the ServiceProvider to the providers array in config/app.php

So it's possible to disable auto-discovery?

barryvdh:
Yes

Issue #700, opened 2017-08-21
Display debugbar only to certain users? L5.4


younus93:
Is there any way that I can set the Debugbar to enabled only for certain users?

aik099:
In your service provider's `boot` method:

1. decide if user can see it
2. do `config(['app.debug' => $condition])` or `config(['debugbar.enabled' => $condition])`

aik099:
Just tried disabling `app.debug` setting from middleware and it had no effect.

aik099:
I think, that we should move code for enabling debug bar into middleware so that anybody could undo enabling instead of enabling right from service provider.

Issue #617, opened 2017-03-15
Enable debugbar in AppServiceProvider


robindirksen1:
I cannot enable the debugbar in the AppServiceProvider when running `\Debugbar::enable();`. Is there another way to use this?

cliftonscott:
Add this inside register() 
`$this->app->register( Debugbar::class );`

Add this at the top
`use Barryvdh\Debugbar\ServiceProvider as Debugbar;`

potsky:
It does not work because in the package boot method, you call :

```
    public function boot()
    {
        ...

        // If enabled is null, set from the app.debug value
        $enabled = $this->app['config']->get('debugbar.enabled');

        if (is_null($enabled)) {
            $enabled = $this->checkAppDebug();
        }

        if (! $enabled) {
            return;
        }
```

So you need to trick in your register method :

```
                $this->app->register(\Barryvdh\Debugbar\ServiceProvider::class);
                $this->app['config']->set('debugbar.enabled',true);
```

potsky:
My trick does not work if environment is `production`...

gsabater:
Is this still true? Since L5.7 this trick is not working for me anymore

MarcosBL:
I was able to get Database queries registered with just

```php
public function register()
{
	$this->app['config']->set('debugbar.enabled', true);
	$this->app->register(\Barryvdh\Debugbar\ServiceProvider::class);
	\Debugbar::enable();
}
```

with only this minor (for me) problem:

```php
Cannot add MailCollector to Laravel Debugbar: Class mailer does not exist
```

ruziwang:
I still have this problem in laravel 6, and I have determined that the laravel debugbar introduces the debug console before routing

Still not displayed

PR #713, opened 2017-09-02
Allow enabling/disabling DebugBar from middlewares

aik099:
What was done:

1. moved route protection code (throws http 500 exception) into base controller constructor (reduced code duplication + auto protect any new controller actions)
2. adjust route protection to also work in Lumen (accessing to `$this->app` causes an error)
3. combined `is enabled` logic from 3 places (`LaravelDebugbar::modifyResponse` method, `LaravelDebugbar::isEnabled` method, `ServiceProvider::boot` method) into `LaravelDebugbar::isEnabled` method
4. moved code, that boots/enables DebugBar into middleware to allow easy manipulation (e.g. enable/disable on the fly) from other middlewares
5. routes are now always registered but protected in case, when accessed on disabled debugbar
6. middleware is now always registered but decides what to do based on enabled state of debugbar

With made changes it's possible to set `debugbar.enabled` or `app.debug` from any of application middlewares and DebugBar's middleware would respect that and enable/disable accordingly.

Should solve usecases from:

* #617 - no tricks, just app middleware
* #700 - app middleware that checks user access level
* #708 - service provider is always loaded, but app middleware would decide if DebugBar actually works

// cc: @robindirksen1, @younus93, @ashgibson


barryvdh:
Should we throw regular 404s instead of 500?

aik099:
> Should we throw regular 404s instead of 500?

Sure. I was actually using 404 at first, but then saw 500 being used and decided to use them instead.

Changing now.

barryvdh:
Yeah but it wasn't registered in the first place before, so was still giving 404

aik099:
Changed.

barryvdh:
Should we use the Middleware function of the controller, if you want to change it on runtime?

aik099:
We probably can create middleware using `$app['debugbar']->isEnabled()` and use `middleware` method on route group in https://github.com/barryvdh/laravel-debugbar/blob/master/src/ServiceProvider.php#L79-L83 to attach it.

Should I do that?

barryvdh:
I meant the Middleware method inside the constructor of the controller

aik099:
I've understood you, but since you have 2 BaseController classes (one for Laravel and other for Lumen), then we can't use `$this->middleware(...` method unfortunately to cover both controllers.

The `->middleware(...` method on route/route group however should work for both Laravel/Lumen (I haven't used Lumen myself).

aik099:
Not sure how it works internally, but maybe defining middleware on route/route group would work faster, because controller class instance isn't even created.

Anyway, should I change anything else?

barryvdh:
defining it on the group would be better perhaps indeed.

aik099:
Done.

barryvdh:
Thanks! Let me know if you run into any problems with disabling now

aik099:
I guess you can safely merge this into 2.4 branch as well to allow Laravel 5.4- users to benefit from this.

barryvdh:
So the reason I didn't do this in the Middleware, is because it's a bit later in the request cycle. Not sure if this is still the best thing, because you could miss some queries/debug messages.

Could we also already boot in the serviceprovider? Downside is that in the event of disabling it on run-time, the listeners would be attached. But imho the useful cases are:

 - Enable on default for certain IP / circumstances (doesn't matter too much if we miss some stuff from early on)
 - Disable to avoid too much output but still in dev, so not extremely important that it's attached, as long as it doesn't output.

aik099:
I've looked at code being executed since service provider boot and middleware execution and haven't found any SQLs being executed (or other things interesting to data collectors). That's why I thought disabling from middlewares is good thing.

PR #728, opened 2017-09-19
feat: add a forget button in the cache panel

Broutard:
Hi,

This PR is a proposal and add a "forget" button to delete a key (it also handle tagged keys).

http://g.recordit.co/vNnbn4wN3Q.gif


barryvdh:
Looks interesting :)

Broutard:
Great.
We could probably find a better way to escape values ( https://github.com/barryvdh/laravel-debugbar/pull/728/commits/651ef9174fc8dfe5bb12f7667716a0a1380f5e3b ), I think you have this issue in others panels.

PR #735, opened 2017-10-08
Use Authorizable in GateCollector

devinfd:
addCheck assumed that the authorized model was the authenticated App\User. In my case I have a App\User, App\Account and an App\AccountUser. The App\User is authenticated but authorizations are on the App\AccountUser which does not implement Authenticatable.

This change type hints $user to Authorizable because thats what is actually relevant in this collector. It also uses the Authorizable's base class name in the message instead of assuming 'user'.

### DIFF
--- a/src/DataCollector/GateCollector.php
+++ b/src/DataCollector/GateCollector.php
@@ -4,7 +4,7 @@ namespace Barryvdh\Debugbar\DataCollector;
 
 use DebugBar\DataCollector\MessagesCollector;
 use Illuminate\Contracts\Auth\Access\Gate;
-use Illuminate\Contracts\Auth\Authenticatable;
+use Illuminate\Contracts\Auth\Access\Authorizable;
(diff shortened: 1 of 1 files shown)


jschutt:
I'm in the same boat here. I'm applying authorizations to a non-authenticable model which is breaking the gate collector. The proposed pull-request fixes the issue for me.

devinfd:
Ping. This is becoming an issue for my team. We can not composed update without this change.

jschutt:
Just checking in to see if this might be approved at some point?

barryvdh:
Is this supported in all recent versions of Laravel?

barryvdh:
Do we need the typehint?

devinfd:
```Illuminate\Contracts\Auth\Access\Authorizable``` is the correct namespace all the way back to laravel 5.1.

I don't really see that the type hint is needed except that you had one originally and this PR is just following that pattern. The typhint could be simplified to just ```Model``` only because of the ```$user->id```

PR #747, opened 2017-11-03
Force injecting with special content types

tianhe1986:
Our project is mainly used as a app server, which returns json format data in most cases.

However,   debugbar would be injected only with `html` content type by default refer [here](https://github.com/barryvdh/laravel-debugbar/blob/master/src/LaravelDebugbar.php#L720). I need to modifying my code to show debugbar.

So I would like to add a config item `force_inject_content_types`,  with which debugbar could be injected with some special content types. 

Signed-off-by: tianhe1986 <w1s2j3229@163.com>

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -179,6 +179,19 @@ return [
 
     /*
      |--------------------------------------------------------------------------
+     | Force Inject Content Types
+     |--------------------------------------------------------------------------
+     |
+     | Debugbar would be injected in Response if the type string appear in the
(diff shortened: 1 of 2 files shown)


barryvdh:
But with json, it would break the valid json?

tianhe1986:
@barryvdh 
Indeed it would break valid json, but also, it would happen when `force_inject_content_types` has been set.
The purpose of this PR is to show the debugbar and help developers debuging with non-html content type. Whether to use it is depend on the developer.

PR #749, opened 2017-11-20
Link the debug bar to the APP_DEBUG variable

andrey-helldar:
Link the debug bar to the APP_DEBUG variable


barryvdh:
We do this in the code.

PR #760, opened 2017-12-24
Loosen maximebf/debugbar constraint

dwightwatson:
This will simply allow a more recent version of the library to be used - `1.15.0`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": ">=7.0",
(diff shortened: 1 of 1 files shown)


barryvdh:
I'm not super comfortable with this, as it might give problems in the future. I'll allow 1.15 though.

barryvdh:
See https://github.com/barryvdh/laravel-debugbar/commit/0b04bcd2264b19317d05b9bb3a059412d6466839

PR #784, opened 2018-02-04
Laravel 5.6 support added

Jurager:
Laravel 5.6 uses Symfony's http-kernel 4.0, exportValue has been deprecated since 3.2 and fully removed in 4.0

PR #787, opened 2018-02-07
Replaced use of removed ValueCollector with VarCloner

A-Lawrence:
As per L5.6, Symfony 4 is now used.  As such, VarCloner must be used in place of ValueCollector.

Builds on commit b5d1df3 & #784

### DIFF
--- a/src/DataCollector/EventCollector.php
+++ b/src/DataCollector/EventCollector.php
@@ -11,7 +11,7 @@ class EventCollector extends TimeDataCollector
     /** @var Dispatcher */
     protected $events;
 
-    /** @var ValueExporter */
+    /** @var VarCloner */
     protected $exporter;
 
     public function __construct($requestStartTime = null)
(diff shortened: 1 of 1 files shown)


lloricode:
hi @barryvdh , can we get this? in a new tag?

Best regards :)

torkiljohnsen:
FYI @barryvdh the upgrade from 3.1.1 to 3.1.2 broke my L5.5 dev env. Would be good with a major version number increment in these cases?

barryvdh:
Well, not it shouldn't break on 5.5. Or are these changes not compatible with Symfony3?

A-Lawrence:
Symfony3 still uses Value Collector I believe.

barryvdh:
But it's deprecated since 3.2 it says in the other issue?

A-Lawrence:
That's a good point - I'd missed that fact.  An error message might be useful.

torkiljohnsen:
Our API returns a 500 now with Debugbar 3.1.2 enabled, both with Laravel 5.5 and Laravel 5.6. Haven't found out why though.

barryvdh:
I've changed it to this due to performance issues: https://github.com/barryvdh/laravel-debugbar/commit/664fa434e26febf04d60df0368303cf0fe14f63a

Should solve the problems still

PR #788, opened 2018-02-10
Laravel 5.6 support, added missed VarColner at EventCollector.php

Jurager:


Jurager:
Oops, it's seems to be already fixed, can you please reject this :)

Issue #776, opened 2018-01-23
QueryCollector still collects queries when the debugbar is disabled


apokryfos:
The issue seems to be at around https://github.com/barryvdh/laravel-debugbar/blob/master/src/LaravelDebugbar.php#L320 

Once the debugbar starts listening to queries it will log all queries regardless of whether it's enabled or not.

This may cause issues (as it has in my case) if there are bulk update tasks which perform a lot of queries with large bindings since they cause a memory leak even if the debugbar is disabled. 

Edit: By memory leak I mean that the request itself along with the bindings is kept in memory for the duration of the request, which is how the debugbar works, but ideally disabling it should prevent it from consuming additional memory.

aleemb:
@apokryfos This issue still persists. Ran into it on the latest release today.

PR #823, opened 2018-04-18
#776 - Check if we're still collecting in the listen handler

apokryfos:
Running `config(["debugbar.collectors.db" => false ]);` should now stop the collection of DB queries until we turn it back on via `config(["debugbar.collectors.db" => true ]);`

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -319,6 +319,9 @@ class LaravelDebugbar extends DebugBar
             try {
                 $db->listen(
                     function ($query, $bindings = null, $time = null, $connectionName = null) use ($db, $queryCollector) {
+                        if ($this->shouldCollect('db', true)) {
+                            return; // Issue 776 : We've turned off collecting after the listener was attached
+                        }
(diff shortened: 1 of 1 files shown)


afraca:
@apokryfos  , is the default (true) ok? I can't check, but I heard from colleague that collection completely stopped for queries...

marceloxp:
In `LaravelDebugbar.php` Line 322, i think is missin a `not if`.

```php
// current code:
if ($this->shouldCollect('db', true)) {
    return; // Issue 776 : We've turned off collecting after the listener was attached
}```

```php
// if NOT shouldCollect => get out :D
if (!$this->shouldCollect('db', true)) {
    return; // Issue 776 : We've turned off collecting after the listener was attached
}```

Thanks ;)

apokryfos:
Yes seems like I pooched it there.

afraca:
It has already been patched here:
https://github.com/barryvdh/laravel-debugbar/commit/d3cdca2ad6cc6e67735b4a63e7551c690a497f5f

PR #841, opened 2018-06-06
Config flag to log request global stats

diogogomeswww:
A config flag that when set to `true` will log something like this:
`[2018-06-06 14:16:08] local.INFO: LaravelDebugbar-RequestStats {"route":"GET embedded/products","duration":"6.41s","memory":"8.94MB","queries":58,"views":93}`

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -163,7 +163,7 @@ return [
             'values' => true // collect cache values
         ],
         'log' => [
-            'request_global_stats' => true // log route, memory_usage, request_duration, number of queries & views
+            'request_global_stats' => false // log route uri, request_duration, memory_usage, number of queries & views
         ]
     ],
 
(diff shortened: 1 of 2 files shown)


afraca:
I understand how this could be useful, but remember this is a debug**bar**  not some generic debugging utility. There are several ways to do what you want. 
* For the route name, you can just ask the router (or use the facade)
* For the duration it's trickier, but you could use middleware for that. Perhaps https://laravel.com/docs/5.5/middleware#terminable-middleware
* If the debugbar is 'running' (for this request will show up), then the query log is enabled. Then you could do `DB::getQueryLog()` in the middelware from my previous point
* The memory usage could also be logged in the middleware with `memory_get_usage`, no magic required
* For the views I have no idea, sorry

diogogomeswww:
I know that this could be done with another external package, which I will do, if this PR is not accepted.

But all the data is already being collected by debugbar, so there's no overhead.
And it's an option, that is set to false by default.

barryvdh:
Thank you, but I think this is out of scope indeed.

PR #846, opened 2018-07-09
Update laravel-debugbar.css

SplashZK:
Change the size of the tables
Before:
antes
After:
despues


barryvdh:
I don't think that's usually the case.. do you have some conflicting cash?

SplashZK:
i don't have css conflict, the css is loaded correctly, when i disable this property the table take the correct form.
image
i don't know if only me have this problem, maybe yes because i search in the issues section and i don't found anything related

diogogomeswww:
I also saw that today. and was about to make a PR.

But I didn't check if that CSS is being used elsewhere.

barryvdh:
Sorry, not really sure about this. Haven't encountered this before.

PR #851, opened 2018-07-19
Enable export all queries to csv

diogogomeswww:
This will add a link to the top-left corner of the queries widget.
image

When clicked all the queries will be copied to the clipboard in csv format.
Allowing the user to paste that data into a spreadsheet and easily: check for duplicates, check for slow queries,....

image

### DIFF
--- a/src/Resources/sqlqueries/widget.js
+++ b/src/Resources/sqlqueries/widget.js
@@ -241,7 +241,10 @@
                     t.append(", " + duplicate + " of which were duplicated");
                     t.append(", " + (data.nb_statements - duplicate) + " unique");
                 }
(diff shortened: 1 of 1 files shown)


barryvdh:
Not 100% how useful this would be for most cases. What would be your usecase?

paezraphael:
I am counting the repeated consultations and sending a report for query analysis.
Making a comparison of the use of the cache.

diogogomeswww:
@barryvdh exporting to excel allows us to easily see the list of queries.
Filter them, counting duplicated, sort by duration,...

barryvdh:
I think Telescope would be easier perhaps, where everything is stored in the database.

PR #866, opened 2018-09-12
Fix style for params table

baorv:
image
Remove: `margin: 10px 100% 10px 0;`


barryvdh:
I think this is probably not relevant anymore, or it's because of conflicting styles

Issue #863, opened 2018-09-04
Add support for Laravel 5.7


lucasdcrk:
Hi, Laravel 5.7 arrived and this package is not compatible (yet) with 5.7 :(

Jeroen-G:
In what way? I'm using it in 5.7

hoRacy:
@Jeroen-G  - when you use gates & policies you have a ReflectionClass error.

bitw:
In config/debugbar.php set false in parameter gate
`'gate'            => false, // Display Laravel Gate checks`

PR #867, opened 2018-09-13
Fix call GateCollector using callback instead of array

### LINKED ISSUE #864: ReflectionFunction::__construct() expects parameter 1 to be string, array given
When using policies in Laravel 5.7 I get the above mentioned error. (Debugbar v3.2)
It also happens on a fresh copy of Laravel.

It's thrown in Illuminate\Auth\Access\Gate.php @ callbackAllowsGuests
Parameter passed to the ReflectionFunction constructor:
array:2 [▼
  0 => GateCollector 
  1 => "addCheck"
]

It happens only when there is no authenticated user.

It occurs when calling the authorize method in the controller:
$this->authorize('action', $model);

And when using the @can blade directive
@can('view', $model)

Douglasdc3:
This fixes #864  & #863

Pre Laravel 5.7 the the `$gate->after` would not explicity check for optional / null. Laravel 5.7 the Gate class in Laravel will explicity check the callback signature this does not work with array arguments.

Gate Class (Laravel 5.7)

```php
protected function callbackAllowsGuests($callback)
{
        $parameters = (new ReflectionFunction($callback))->getParameters();

        return isset($parameters[0]) && $this->parameterAllowsGuests($parameters[0]);
}
```

The `$callable` will be array  as argument, The `ReflectionFunction` can only take callable argument.

The reason we get this far into the code and is not cuaght by the Gate contract is the fact that the contract expects a `callable` type since an array is considered callable if you use a object reference as the first argument and string argument as second argument as method name.

This is a bit of breaking change between Laravel 5.6 - Laravel 5.6. My fix is backwards compatible with Laravel 5.6.

### DIFF
--- a/src/DataCollector/GateCollector.php
+++ b/src/DataCollector/GateCollector.php
@@ -20,7 +20,9 @@ class GateCollector extends MessagesCollector
     {
         parent::__construct('gate');
         $this->setDataFormatter(new SimpleFormatter());
-        $gate->after([$this, 'addCheck']);
+        $gate->after(function (Authenticatable $user = null, $ability, $result, $arguments = []) {
+            $this->addCheck($user, $ability, $result, $arguments);
+        });
     }
 
     public function addCheck(Authenticatable $user = null, $ability, $result, $arguments = [])


drbyte:
Thanks @Douglasdc3 -- That fixed it for me!

ali-ann:
Works like a charm. Thanks @Douglasdc3

Issue #363, opened 2015-06-19
Small fix for the EXPLAIN regex


crosma:
I didn't think this warranted a pull (also, I've never done one, and I'm lazy). The implode and regex both needed small fixes. The \s\* will allow for any amount of whitespace before the keyword for those who hand write their queries and format them.

QueryCollector.php:93

```
if ($this->explainQuery && preg_match('/^\s*('.implode('|', $this->explainTypes).')/i', $query)) {
```

PR #876, opened 2018-10-09
Fix explain mode regex #363

hrs-o:
I am in trouble with the problem that occurred at #363.

When multiple explainTypes were specified, corrected problem which can not be judged correctly.


barryvdh:
According to https://github.com/barryvdh/laravel-debugbar/issues/888 this creates a problem. Could you check that out?

barryvdh:
A wait, that the Update/delete/insert actually never properly worked before this fix. But now it's merged, we entounter the issues. So this PR is actually correct, only it brings out the incorrect behavior. Only SELECT is safe to run like this.

PR #883, opened 2018-10-23
Telescope compatibility 

barryvdh:
- Disable Telescope on Debugbar routes
- Disable Debugbar op Telescope routes

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -16,7 +16,7 @@ return [
 
     'enabled' => env('DEBUGBAR_ENABLED', null),
     'except' => [
-        //
+        'telescope*'
     ],
 
     /*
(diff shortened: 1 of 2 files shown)


barryvdh:
Yes it will not throw an error, because of the class_exists check.

johanvanhelden:
@barryvdh when are you planning to release this?

barryvdh:
Not really sure because Telescope is still in beta I think. But probably could be released already. The config is something everyone can add already.

barryvdh:
@themsaid Do you think this is the most reliable way to disable Telescope, or could debugbar be ignored on the side of Telescope? Or a config flag?

johanvanhelden:
I've taken a look at the Telescope source, but it does not look like it has a `serving` method like Nova does. So that makes disabling the Debugbar from Telescope a bit difficult. Nor does it seem to have a blacklist like Debugbar.

PR #886, opened 2018-11-02
\Exception isn't well with newer PHP

roman-curse:
Propose to use Throwable instead


Cannonb4ll:
What is actually the difference between Exception & Throwable?

barryvdh:
Internal errors are 'Error'. Both Error and Exception extend Throwable. So Throwable is wider

PR #906, opened 2019-02-15
Dont register routes and middleware if we are not in debug mode

vkarampinis:
ServiceProvider as is now, register routes and middleware even when we are not on debug mode `APP_DEBUG=false`

This patch allows ServiceProvider to boot only when `app.debug` is `TRUE`

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -63,6 +63,10 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
      */
     public function boot()
     {
+        if (!$this->app['config']->get('app.debug')) {
+            return;
+        }
+        
         $configPath = __DIR__ . '/../config/debugbar.php';
         $this->publishes([$configPath => $this->getConfigPath()], 'config');
 


JaZo:
We are using the debugbar in a staging environment without `APP_DEBUG=true` (we don't want to see a detailed exception page etc.) and only enable it deferred when a user is logged in. Although the docs state `APP_DEBUG=true` is required, it has been working just fine for years. Unfortunately this change broke it and the only fix is to overwrite the boot method of the ServiceProvider. 

I've been thinking about making a PR with something like [LaravelDebugbar::isEnabled](https://github.com/barryvdh/laravel-debugbar/blob/master/src/LaravelDebugbar.php#L751) to check `config('debugbar.enabled')` and fallback to `config('app.debug')`, but that doesn't work if we enable the debugbar deferred 😞 @barryvdh, do you have an idea if and how we can support our use case?

vkarampinis:
Good catch.
replacing this line will do the trick.
https://github.com/barryvdh/laravel-debugbar/blob/master/src/ServiceProvider.php#L66

```diff
- if (!$this->app['config']->get('app.debug')) {
+ if (!app('debugbar')->isEnabled()) {
```

@barryvdh your call. I can send a PR for this.

JaZo:
@vkarampinis, unfortunately that would break [enabling it at runtime](https://github.com/barryvdh/laravel-debugbar#enablingdisabling-on-run-time).

aalyusuf:
This also renders config's "enabled" obsolete as it cannot actually be overriden.

barryvdh:
Should we just revert this?

JaZo:
@barryvdh, I think it's better to do so. Registering those routes and the middleware doesn't look like a big performance issue so I don't see a reason why we shouldn't register those.

barryvdh:
Reverted

mnabialek:
@barryvdh  Please make new release with this fix. I'm using 3.2.3 and it's not reverted in this release

barryvdh:
done

PR #910, opened 2019-02-26
Avoid to hook debugbar in response if its an ajax request

vkarampinis:
InjectDebugbar middleware modifies response without detecting if its an ajax call or not.

There are cases where we return views via ajax call for templating (Angular etc).
Injecting Debugbar on that response is useless.

This patch checks if its an`isXmlHttpRequest` before inject Debugbar

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -723,7 +723,7 @@ class LaravelDebugbar extends DebugBar
                 strpos($response->headers->get('Content-Type'), 'html') === false)
             || $request->getRequestFormat() !== 'html'
             || $response->getContent() === false
-            || $request->isXmlHttpRequest()
+            || $this->isJsonRequest($request)
         ) {
             try {
                 // Just collect + store data, don't inject it.
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -723,6 +723,7 @@ class LaravelDebugbar extends DebugBar
                 strpos($response->headers->get('Content-Type'), 'html') === false)
             || $request->getRequestFormat() !== 'html'
             || $response->getContent() === false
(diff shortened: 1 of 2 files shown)


barryvdh:
We actually check in modifyResponse if it's an json request, we should probably check for Ajax there also. We do want to inject headers etc, because we can track Ajax calls.

vkarampinis:
You have right. It will be better to check inside `modifyResponse` for ajax calls.
I'll try to send a fix.

vkarampinis:
I move the check inside `modifyResponse`.
Let me know if you think this is ok.

src/LaravelDebugbar.php
```diff
@@ -723,6 +723,7 @@ public function modifyResponse(Request $request, Response $response)
                 strpos($response->headers->get('Content-Type'), 'html') === false)
             || $request->getRequestFormat() !== 'html'
             || $response->getContent() === false
+            || $request->isXmlHttpRequest()
```
barryvdh: Can you use `$this->isJsonRequest($request)` like above?

barryvdh:
Do you have capture ajax disabled? Because `$this->isJsonRequest($request)` does the same check already. But when it's disabled, it should indeed not try to inject it.

vkarampinis:
Yes, I have `'capture_ajax' => false`, if it's enabled then Debugbar don't hook in response and capture data with ajax call. As expected.

src/LaravelDebugbar.php
vkarampinis: sure

barryvdh:
Thanks! But still, it should be enabled, otherwise I don't think Debugbar will capture anything?

besrabasant:
Hi @barryvdh @vkarampinis 

I checked with this patch and the issue still remains.

PR #918, opened 2019-03-25
Revert "Dont register routes and middleware if we are not in debug mode"

barryvdh:
Reverts barryvdh/laravel-debugbar#906


stayallive:
@barryvdh It would highly appreciate it if you can find the time to tag this in a release to prevent us needing to pin the dependency 😄

PR #931, opened 2019-05-30
Fix broken table representation in Views

FallenBG:
under Chrome 74.0.3729.157 the params table in view tab is broken.
It seems that changing the margin to 60% doesn't affect Firefox and IE(where it worked with 100%) but fixes Chrome.

Before:
params_before

After:
params_fix


barryvdh:
I think this is not relevant anymore, or at least I haven't encountered it.

PR #932, opened 2019-06-03
Fixed #306

### LINKED ISSUE #306: Exception Call to a member function quote() on null when saving records
The debug toolbar works as expected when loading pages and browsing.

However when I try to save a record to the database, I get a:

```
'Symfony\Component\Debug\Exception\FatalErrorException' with message 'Call to a member function quote() on null'
```

The error suggests it is coming from: 

```
vendor/barryvdh/laravel-debugbar/src/DataCollector/QueryCollector.php:102
```

When I disable the toolbar with config.php, records are saved as expected with no errors.

Anyone else experiencing this issue? Any suggestions?

unti1x:
QueryCollector now skips the pdo-based operations for the engines that don't use it. Fixed #306

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -78,22 +78,13 @@ class QueryCollector extends PDOCollector
 
     /**
      *
-     * @param string $query
+     * @param \PDO $pdo
      * @param array $bindings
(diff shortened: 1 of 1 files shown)


barryvdh:
Not sure if this is still relevant.

PR #933, opened 2019-06-04
Fix DocBlock: `message` parameter type

MohannadNaj:
Hello, Thanks for this awesome project!

On the Facade the `$message` parameter is typehinted as a string, while it should be mixed ([A message can be anything from an object to a string](https://github.com/barryvdh/laravel-debugbar/blob/2d195779ea4f809f69764a795e2ec371dbb76a96/src/LaravelDebugbar.php#L969)).

Discovered this when using a static analysis tool, (I used [larastan](https://github.com/nunomaduro/larastan)), an error message like this will be collected:
```
Parameter #1 $message of static method Barryvdh\Debugbar\Facade::info() expects string, array<string, string> given.
```


barryvdh:
Thanks!

PR #944, opened 2019-07-23
Allow additional excluded paths for db backtrace

zhuston:
Our project uses a custom database connection and it was always listed as the file source in the backtrace. Adding this feature allowed us to make use of the very helpful backtraces.


joeyrush:
+1

This is useful when using packages that intercept queries (such as model caching). In my case I'm applying a vendor trait to my base model and so every query in my debugbar is showing a vendor path, which I really don't care about

PR #947, opened 2019-07-29
Add new models collector

reinink:
This PR adds a new "models" collector. It essentially listens for `eloquent.retrieved` events, and totals them by model type. This provides a super helpful insight into how many Eloquent models are being generated for a request...something that can have a huge impact on performance.

At first I had it where you could click on each model type to view all the loaded models. However, after chatting with @barryvdh, we decided it was probably better to just show the totals, to keep things snappy.

image

By default this metric is disabled, but I could honestly see this being enabled by default in the future, if folks find it as useful as I do.

### DIFF
--- a/src/DataCollector/ModelsCollector.php
+++ b/src/DataCollector/ModelsCollector.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Barryvdh\Debugbar\DataCollector;
+
+use Barryvdh\Debugbar\DataFormatter\SimpleFormatter;
+use DebugBar\DataCollector\MessagesCollector;
+use Illuminate\Events\Dispatcher;
+use Illuminate\Support\Str;
+
+/**
+ * Collector for Models.
+ */
+class ModelsCollector extends MessagesCollector
+{
+    public $models = [];
+
+    /**
+     * @param Dispatcher $events
+     */
+    public function __construct(Dispatcher $events)
+    {
+        parent::__construct('models');
+        $this->setDataFormatter(new SimpleFormatter());
+
+        $events->listen('eloquent.*', function ($event, $models) {
+            if (Str::contains($event, 'eloquent.retrieved')) {
+                foreach ($models as $model) {
+                    $class = get_class($model);
+                    $this->models[$class] = ($this->models[$class] ?? 0) + 1;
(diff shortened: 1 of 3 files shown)


barryvdh:
Looks good, gonna try this out for a bit :)

reinink:
Awesome! I'll do the same. Thanks Barry! ❤️

src/DataCollector/ModelsCollector.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    public $models = [];
+
+    /**
+     * @param Dispatcher $events
+     */
+    public function __construct(Dispatcher $events)
+    {
+        parent::__construct('models');
+        $this->setDataFormatter(new SimpleFormatter());
```
juancho48: is this being used somewhere else?

reinink: Yes...I stole it from other collectors. For example:

https://github.com/barryvdh/laravel-debugbar/blob/bda348f9105a3cdd6dd9bd3966298b69fdd1dfcc/src/DataCollector/GateCollector.php#L19-L22

kbussche:
Any ETA on when this will get tagged (3.2.5) ?

emargareten:
@reinink can this be added to Telescope as well?

src/LaravelDebugbar.php
```diff
@@ -404,6 +405,15 @@ function ($query, $bindings = null, $time = null, $connectionName = null) use ($
             }
         }
 
+        if ($this->shouldCollect('models', false)) {
```
latheesan-k: The model collector only runs if the flag is turned off?

Shouldn't this be `if ($this->shouldCollect('models', true)) {`

barryvdh: no

jethromay:
Any word on when this is being released @barryvdh ? I would love to make use of this in a current project.

barryvdh:
Yes 3.2.5 was released August 21. See https://github.com/barryvdh/laravel-debugbar/releases

MrEduar:
Great, I didn't know this existed until I saw your videos ❤️
And since I still had an old version these were not shown by default.

Issue #856, opened 2018-08-03
[3.1.5] Debugbar prevents Session from persisting after it expired


MatusMak:
I have encountered an issue that after much investigation seems to be linked to Debugbar. In certain case, it prevents Session from re-establishing after it was expired. I reproduced this regardless of PHP version, OS or browser. Below I hopefully gave you all necessary information, if you need anything else feel free to ask for it :).

Here are few scenarios:
-----------------------

1. If no user is logged in, everything works fine.

2. If admin user is logged in (with "Remember me" being checked), but public user is not, accessing any public (unprotected) route (which uses default `web` middleware) will recreate session as described above. However, once I access any route under `admin` middlware (`web` middleware + auth check), session will stick and after that, the issue for public route disappears.

3. If both users are logged in and I access any protected route, session sticks immediately.

What resolves the issue:
------------------------

 - accessing protected page
 - deleting cookies
 - restarting browser

Steps to reproduce (ONLINE):
---------------------------------

On this environment, SESSION_LIFETIME was set to 1 so it can be quickly reproduced. Environment is set to local and debug is set to true.
 - Go to http://laravel-issue-reproduction.6f.sk/
 - Open debug bar. Notice the _token value (XSRF token is tied with session - once session regenerates, it will regenerated as well. If you do not trust that new session is created, I can give you FTP access and you can observe creation of new session files in storage folder yourself - just let me know :) ).
 - Go to http://laravel-issue-reproduction.6f.sk/admin (preferably on a new tab)
 - Log in and **check remember me** (credentials are displayed here)
 - Now observe that both tokens are the same, as they should be.
 - Wait 2 minutes and do not refresh anything.
 - Refresh http://laravel-issue-reproduction.6f.sk/. Observe that _token changed, as it should since session expired.
 - Refresh multiple times. Observe that _token changes with each refresh.
 - Refresh http://laravel-issue-reproduction.6f.sk/admin. Observe that _token changed.
 - Refresh /admin - token sticks
 - Refresh / - token now sticks and is same as on /admin

You can download this demo [here from BitBucket](https://bitbucket.org/MatusMak/laravel-issue-repro/src/master/)
If you would like to have FTP access, I can give you one.

----------

Reference to issue opened (and closed) in Laravel project: https://github.com/laravel/framework/issues/25081

MatusMak:
I have taken down site I originally set up for reproduction. Please refer to ticket I opened in laravel/framework for more information on the issue.

[...]

PR #948, opened 2019-08-02
Fix session destroy when using multiple guards

skyosev:
Related to #856 
It seems the recaller logic is outdated as it leads to session destroy (and thus csrf token regeneration) when using multi-auth. It fails to retrieve the guard from the cookie, which leads to $guard->user() call and session migration.


barryvdh:
Does this require to bump the minimum version?

skyosev:
I think no. The 'Recaller' logic is available since Laravel 5.4.

barryvdh:
I was still getting some errors because Provider doesn't exist (and you didn't import Recaller).
Can you check 949?

PR #962, opened 2019-08-27
Protect against possible null values in `$models` array.

mikebronner:
I came across this error today after enabling models collector. (Thanks btw for this awesome feature @reinink!):

```
get_class() expects parameter 1 to be object, null given (View: /private/var/tmp/homepage-articles.indexmKf4EN)
```

Screen Shot 2019-08-27 at 8 09 41 AM


barryvdh:
Thanks!

PR #965, opened 2019-09-03
Fixed get_class() for empty model

rokde:
In Laravel 5.7.x it is possible the event retrieves an empty model - so the `get_class()` call throws an `ErrorException`. By catching this it stays working, but it does not count the try of retrieving.

### DIFF
--- a/src/DataCollector/ModelsCollector.php
+++ b/src/DataCollector/ModelsCollector.php
@@ -25,8 +25,12 @@ class ModelsCollector extends MessagesCollector
         $events->listen('eloquent.*', function ($event, $models) {
             if (Str::contains($event, 'eloquent.retrieved')) {
                 foreach (array_filter($models) as $model) {
-                    $class = get_class($model);
-                    $this->models[$class] = ($this->models[$class] ?? 0) + 1;
+                    try {
+                        $class = get_class($model);
+                        $this->models[$class] = ($this->models[$class] ?? 0) + 1;
+                    } catch (\ErrorException $exception) {
+                        // eloquent model retrieved can be null
+                    }
                 }
             }
         });


Tegos:
Hello @rokde 
Is it fresh installation of 5.7? If it simple empty model it's already fixed.

I discovered similar problem but is tied with compatibility. I used package:
https://github.com/GeneaLabs/laravel-model-caching and it used https://github.com/GeneaLabs/laravel-pivot-events as dependency.
It affects original events for eloquent: 
https://github.com/GeneaLabs/laravel-pivot-events/blob/master/src/Traits/ExtendFireModelEventTrait.php#L36

image

In ModelsCollector, I noticed that now exists array_filter in foreach, it will fixed fatal error on get_class on null, but in this case it will do double counting (array keys 0 and 'model').

So my possible fix:
```
foreach (array_filter($models) as $key => $model) {
	if (is_numeric($key)) {
		$class = get_class($model);
		$this->models[$class] = ($this->models[$class] ?? 0) + 1;
	}
}
```
What could you tell? @barryvdh @mikebronner

rokde:
It is a nearly fresh install with new Model and test a route binding with an not-existing model. The $model variable is `null` in my case.

rokde:
@Tegos On another project using the `laravel-pivot-events` package I get the same error.

mikebronner:
Isn't this the same that has already been addresses in my PR: https://github.com/barryvdh/laravel-debugbar/pull/962? (not being sarcastic, please let me know if I'm wrong and what is different)

rokde:
@mikebronner yes, you are right. Your change is not in a release so I did not see it. I close this pull request.

mikebronner:
@rokde I pulled in dev-master as a work-around until it gets released:
```sh
composer require "barryvdh/laravel-debugbar:dev-master@dev"
```

barryvdh:
Tagged

Tegos:
Thank you guys. But what about doubled model's count?

See simple example:
```
Route::get('/debugbar', function () {
	$currency = Currency::query()->first();
	dump($currency);
});
```

image

@mikebronner, I guess so you used laravel-cache with laravel-pivot-events, could you test it on your project, please?

PR #985, opened 2019-11-12
feat: add configuration of additional route group middleware

DiskoPete:
Hi!

Thanks for your awesome work on the DebugBar!\
I found no guidelines how to prepare a pull request. So please bare with me :)

For a project I am working on, I needed specific middleware run on every request.\
So I created the new configuration option `route_additional_middleware`.

The additional middleware is then appended to the existing `DebugbarEnabled` middleware.

Thanks for your time!

Best,
Peter

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -199,4 +199,13 @@ return [
      | To override default domain, specify it as a non-empty value.
      */
     'route_domain' => null,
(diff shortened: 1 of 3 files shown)


barryvdh:
I think this is a bit too specific for your use-case

PR #986, opened 2019-11-18
Stop disabling 'testing' environment

dfoxx:
Optionally, if you prefer, I can create another config variable

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -769,7 +769,7 @@ class LaravelDebugbar extends DebugBar
                 $configEnabled = $config->get('app.debug');
             }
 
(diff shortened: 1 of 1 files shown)


barryvdh:
Why do you want it enabled in testing?

dfoxx:
I hadn't considered phpunit, but I understand now why this is in there. My issue was that I was expecting to see the bar if I set `APP_DEBUG=true` or `DEBUGBAR_ENABLED=true` in the `.env`. If this is off the table to save headaches with phpunit, I will edit my request to update the readme disclaiming this exception.

barryvdh:
I don't really understand why Debugbar should be visible with testing. It is a development tool only and will drastically slow down any tests otherwise.

PR #995, opened 2019-12-10
Added numerical ordering for widgets list

duysolo:
Added numerical ordering for widgets list, such as: 
- Views list 
- SQL queries list

Capture

### DIFF
--- a/src/Resources/laravel-debugbar.css
+++ b/src/Resources/laravel-debugbar.css
@@ -204,6 +204,10 @@ div.phpdebugbar-widgets-messages div.phpdebugbar-widgets-toolbar a.phpdebugbar-w
     text-align: center;
 }
 
+ul.phpdebugbar-widgets-list {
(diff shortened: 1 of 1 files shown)


duyphan2502:
@barryvdh Could you have a look on it? 
The numerical ordering is very useful for us when we have a long widgets list.

xuandung38:
Hope approve ... soon

sangnguyenplus:
Waiting for approval 👍

ManojKiranA:
Adding this to Config file may be Helpful too.

I think You can add that under Config->options array

So that Users may Enable or Disable it

sangnguyenplus:
@ManojKiranA I don't think so, it's just a small feature with some extra CSS. Don't need a configure for a simple feature like that.

barryvdh:
What do those numbers add? You can't connect them to any index or anything?

PR #1006, opened 2020-02-14
fix error Undefined property SessionGuard

KarimQaderi:
```
Undefined property: Illuminate\Auth\SessionGuard::$id
```


barryvdh:
How can I reproduce this?

KarimQaderi:
I do not remember how this happened

PR #1014, opened 2020-03-03
Add timing information to events

### LINKED ISSUE #390: Events execution time is always 0μs.
I'm pretty sure that it is not correct. I've found this in onWildcardEvent method:

```
$time = microtime(true);
$this->addMeasure($name, $time, $time, $this->prepareParams($args) );
```

You can see that start and stop times are the same.

LukeTowers:
This fixes #390 and expands on #392. Additionally, the change to referencing the Dispatcher contract rather than the implementation allows for greater flexibility (specifically I'm using it with October CMS, which has its own event dispatcher that implements the Dispatcher contract).

While the timing information isn't 100% accurate, as the first event will be tracked if it took all of the time from when the collector was setup to when the event fired, and any future events are really just tracking the time between itself and the previous event, this is still valuable information that is close enough to reality to act upon. Additionally, this collector was setup from the start to track event timings, but was never actually implemented properly as it would record the start and end of the event as being the same time, always resulting in a timing of 0.

### DIFF
--- a/src/DataCollector/EventCollector.php
+++ b/src/DataCollector/EventCollector.php
@@ -3,7 +3,7 @@ namespace Barryvdh\Debugbar\DataCollector;
 
 use Barryvdh\Debugbar\DataFormatter\SimpleFormatter;
 use DebugBar\DataCollector\TimeDataCollector;
-use Illuminate\Events\Dispatcher;
+use Illuminate\Contracts\Events\Dispatcher;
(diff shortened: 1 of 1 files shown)


LukeTowers:
@barryvdh @GrahamCampbell could one of you review this and let me know your thoughts?

LukeTowers:
Thanks @barryvdh, I appreciate the review!

PR #1020, opened 2020-04-13
Complete redesign of the DebugBar (CSS only)

TristanMouchet:
Hey,

I was finding the bar messy and not very intuitive, this CSS update is improving [a lot] the overall look of the bar and its features.

Hope you like it!

### DIFF
--- a/src/Resources/laravel-debugbar.css
+++ b/src/Resources/laravel-debugbar.css
@@ -434,14 +434,15 @@ a.phpdebugbar-maximize-btn {
 }
 
 a.phpdebugbar-tab.phpdebugbar-active {
-    background: #fa262bbf;
+    background: #fa5661;
+    background-image: none;
     color: #fff;
     text-shadow: 1px 1px #000;
 }
 
 a.phpdebugbar-tab.phpdebugbar-active span.phpdebugbar-badge {
     background-color: white;
-    color: #fa262bbf;
+    color: #fa5661;
     text-shadow: 1px 1px #ebebeb;
 }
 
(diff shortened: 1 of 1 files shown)


barryvdh:
Thanks, do you have a screenshot?

TristanMouchet:
Here are some screenshots. :)
debug-bar-1
debug-bar-2
debug-bar-3
debug-bar-4
debug-bar-5
debug-bar-6
debug-bar-7
debug-bar-8
debug-bar-9

I saw there are some issues with Internet Explorer, I will check what's wrong.

TristanMouchet:
Internet Explorer issues are due to a JavaScript error on the original phpdebugbar, so once this is fixed, it should be alright.

https://github.com/maximebf/php-debugbar/pull/441

barryvdh:
Thanks! Overall I like this, a lot cleaner :)

Few points I'm not sure about:
 - Why is there a grey line above the red border. Is that by design still?
image

- I'm not sure if I'm a big fan of the 'well' like header, I think that it look cleaner before (and more minimal).
image

- I see you made the tables/rows more consistent, but I think without the borders, we might need to make the contrast a bit bigger between odd/even? What do you think? (Same goes perhaps for the hovers)
image

But overall very happy with this :) Let me know if you can do something with these suggestions, or otherwise I'll take a stab at it myself.

TristanMouchet:
Here we go:

debug-bar-10

debug-bar-11

barryvdh:
Ah, nice! The things I pointed out look much better now, thanks!

Is this good to merge?

TristanMouchet:
Yep all good now. :)

barryvdh:
FYI, feedback overall is positive :)
https://twitter.com/barryvdh/status/1250695650548895745

Just tweaked some small things with some custom collectors and text-shadows.

PR #1023, opened 2020-04-16
Enabled models collector by default

reinink:
Back in July of 2019 we added the models collector (#947). I have found this to be a *very* useful collector on my Laravel projects, since models are such a key part of the framework. These model metrics give insight into performance issues that's not always obvious from the queries or memory usage.

I'd suggest that we enable this by default. 🙏


reinink:
🕺🥳🎉

PR #1037, opened 2020-05-11
Utilize the screen space for easier reading

mlantz:
Most developers when working have screens that well exceed 200px. Given that many models, route middlewares, session names etc can exceed 200px in length. It seems reasonable to give them space for readability.


reinink:
Looks too to me! 👍 

image

PR #1050, opened 2020-06-16
Move route definition into separate file

fletch3555:
I have a use-case where I need to flush and reload different routes to change context. When doing this, you lose any routes that can't be manually reloaded.  Since these routes are defined in the provider itself, I have no way to re-register them.

As a solution, I've moved the routes into their own file.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -26,6 +26,8 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
     {
         $configPath = __DIR__ . '/../config/debugbar.php';
         $this->mergeConfigFrom($configPath, 'debugbar');
+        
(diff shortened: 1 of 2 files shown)


fletch3555:
@barryvdh, I'm sure you're incredibly busy, but have you had a chance to look at this? Anything I can do to help move it along?

andersLAL:
This seems to cause a ReflectionException, Class DebugbarEnabled does not exist, since DebugbarEnabled isn't imported in the new debugbar-routes.php file.

fletch3555:
Ah, it would seem you're right.  Terribly sorry for that.  I swear I tested it, so no clue how I missed it.

You found it, so would you like to PR the fix, or shall I?

andersLAL:
No worries :) I'll just leave it to you if you don't mind.

barryvdh:
Reverted

PR #1051, opened 2020-06-17
Optimize models collector

sebdesign:
Use a wildcard to listen to the `eloquent.retrieved:*` events instead of listening to all the `eloquent.*` events.

Also I took the liberty to add some tests! 😅


barryvdh:
Nice. Was just thinking I should add some tests :p

PR #1055, opened 2020-06-27
Move route definition into separate file - v2

fletch3555:
Fixed issue reported in #1050


barryvdh:
So did you test this?

fletch3555:
Quick response...

Apologies, I had missed something.  The use of $this->app() does not work in certain circumstances.  Do you prefer using facades (`Config`, `Route`) or helper methods (`config(...)` or `app('config')`, `app('route')`)?

fletch3555:
Just following up on this.  Any opinions on my question, @barryvdh ?

fletch3555:
@barryvdh thanks for the merge, but I had an outstanding question.  Just curious if you prefer `config(...)` or `app('config')`?  There's one scenario where this won't work as-is.  I'll open another PR with the fix,

PR #1056, opened 2020-07-01
Adds check for availability of jQuery

niveshsaharan:
Fixes Uncaught TypeError: Cannot read property 'noConflict' of undefined when jQuery is not yet initialized.


barryvdh:
Not really sure why jQuery wouldn't be intialized.

niveshsaharan:
I’m not sure either. I am only using Laravel Livewire in my app and it always throws error in console.

barryvdh:
Are you using Turbolinks? I think that should be fixed with https://github.com/barryvdh/laravel-debugbar/pull/1084

niveshsaharan:
Thank you @barryvdh 

yes, I use Turbolinks. I will check it.

PR #1060, opened 2020-07-16
[queries] Support double question mark

cmizzi:
As you may know, since PHP 7.4, it is now possible to use special operators (known mainly under PostgreSQL as `?`, `?|`, `?&`...) by doubling the question mark (`??`, `??|`, `??&`...). The formatter didn't support this feature, it's done now.

In addition, the formatter replaces the question mark escape so that it can be easily copied and pasted to debug a query. I also implemented a test on this feature.

```sql
SELECT ('[1, 2, 3]'::jsonb ?? ?) as a, ('[4, 5, 6]'::jsonb ??| ?) as b, 'hello world ? example ??' as c

Bindings : [3, '{4}']
```
```diff
-SELECT ('[1, 2, 3]'::jsonb 3'{4}' ?) as a, ('[4, 5, 6]'::jsonb ??| ?) as b, 'hello world ? example ??' as c
+SELECT ('[1, 2, 3]'::jsonb ? 3) as a, ('[4, 5, 6]'::jsonb ?| '{4}') as b, 'hello world ? example ??' as c
```

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -126,7 +126,6 @@ class QueryCollector extends PDOCollector
                     ? "/(?<!\?)\?(?=(?:[^'\\\']*'[^'\\']*')*[^'\\\']*$)(?!\?)/"
                     : "/:{$key}(?=(?:[^'\\\']*'[^'\\\']*')*[^'\\\']*$)/";
 
-
                 // Mimic bindValue and only quote non-integer and non-float data types
                 if (!is_int($binding) && !is_float($binding)) {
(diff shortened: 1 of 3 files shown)


barryvdh:
How does this work? Is that postgre specific?

cmizzi:
@barryvdh 

At the moment, it seems that only PostgreSQL defines operators containing `?` (not enough skill on other DBs to confirm). However, the RFC has been directly implemented in SQL Parser PDO.

- https://wiki.php.net/rfc/pdo_escape_placeholders ;
- https://github.com/php/php-src/pull/4217 ;

PR #1068, opened 2020-07-29
Add fallback string quoting for PDO drivers that don't support it and fix the FE support for named parameter bindings.

ic1011:
Some PDO drivers do not support string quoting for use in SQL queries and the PDO->quote method returns False in these cases. This is the case for DB2 ODBC for IBM i series. Fixed by providing simple fallback quoting.

The FE bindings metadata section under the Queries tab isn't rendered for SQL queries where named parameter placeholders are used because the logic always expects data of type Array. Fixed by creating an array from the object's properties values.

Fix for #1070 and #1071


barryvdh:
I'm not really sure, I think we use the same logic as Laravel does, don't we? Or how does quoting work in Laravel itself?

ic1011:
> 
> 
> I'm not really sure, I think we use the same logic as Laravel does, don't we? Or how does quoting work in Laravel itself?

Sorry for the late answer. Laravel uses value binding and prepared statements instead of quoting and variable interpolation for database queries. Also, currently the related debugbar logic uses directly PHP's PDO class and effectively is skipping any added logic by Laravel: https://github.com/barryvdh/laravel-debugbar/blob/233c10688f4c1a6e66ed2ef123038b1363d1bedc/src/DataCollector/QueryCollector.php#L122

bennettblack:
@ic1011 Also encountered this issue with DB2 for IBM iSeries. However, using the fallback quoting didn't seem to fix the issue for me.

ic1011:
@bennettblack, I've concluded my work with the project /where the fix was needed/ more than a year ago but my colleagues are still using it and it works for them.

Could it be that some cache /opcache or similar/ is playing tricks on you?

Though I am a bit rusty on the subject as I haven't dealt with it recently, I may be able to help you out if you tell me what you have done so far.

ic1011:
I am not sure why the ternary doesn't work for you. According to the PHP manual:

> https://www.php.net/manual/en/pdo.quote.php: Returns false if the driver does not support quoting in this way.

Maybe you're using different driver that is returning not equal to false value. We're using iSeries's ODBC drivers . But anyway. I am glad you got it working!

Regarding the adapter. Yes, we are using the same package. And yes, my impressions are that only fairly large organizations use DB2-based solutions and usually they don't use Laravel.

barryvdh:
Sorry doesn't really feel right.

PR #1074, opened 2020-08-01
Add dark mode

ekvedaras:
Colors used: https://tailwindcss.com/docs/customizing-colors/#default-color-palette

Screenshot 2020-08-01 at 19 39 10
Screenshot 2020-08-01 at 19 39 04
Screenshot 2020-08-01 at 19 37 51
Screenshot 2020-08-01 at 19 38 03
Screenshot 2020-08-01 at 19 36 03
Screenshot 2020-08-01 at 19 36 50
Screenshot 2020-08-01 at 19 37 07


ekvedaras:
Maybe it should be made optional in case some people want to force specific version for any reason? There could be a config option something like this `'theme' => 'auto', // auto, dark, light`

barryvdh:
I like this idea. I would prefer it to be optional indeed, perhaps auto is a good default.

ekvedaras:
So this change required to duplicate css for dark mode. One for `.dark` and one for `.dark-auto`. I tried to optimize the CSS as much as possible but it is still quite some lines.

There is another way to do it though. We could keep only `.dark` variant in CSS (so no duplicates) and use JS to detect dark/ligt mode changes and apply/remove `.dark` class. If you prefer that option I can push it.

If there is a better way to add classes to debugbar please tell me and I will update :)

ekvedaras:
Here is the other option https://github.com/barryvdh/laravel-debugbar/compare/master...ekvedaras:dark-mode-js

However, both options have one small drawback. Since the classes are added with javascript and it is done after debug bar is initialized, sometimes you can see the light version for a split second. I don't know if there is a way to do that before it is rendered. Maybe modify debug bar content itself before attaching to the request 🤔

ekvedaras:
Or.. we may have separate CSS files for dark mode and add them conditionally. Just then again duplicate versions are needed. Unless, https://www.w3schools.com/cssref/pr_import_rule.asp this is a thing and it works in plain css :D

ekvedaras:
OK, so I've pushed another version of it. This one has no problems with flickering modes on page refresh and only has one CSS file for dark mode to maintain. The only hacky thing was media start and end files, but I think it's a pretty small price to pay.

barryvdh:
I added the Dracula theme for HighlightJS, I think that's a bit more readable/easier to the eyes. What do you think?

image

sebdesign:
The duplicate SQL queries background color needs to be inverted too.

image

barryvdh:
Can either of you submit a PR?

sebdesign:
I'm at work now so I can't work on this until the evening.

barryvdh:
Something like this?
image

barryvdh:
Or more yellow/brownish perhaps:
image

ekvedaras:
Since it's meant to indicate duplicates, I think the yellow one grabs attention better

ekvedaras:
I guess for this color, widget text should be adjusted 
image

ekvedaras:
Gray 500 looks better:

image


```css
div.phpdebugbar div.phpdebugbar-widgets-exceptions li.phpdebugbar-widgets-list-item.phpdebugbar-widgets-sql-duplicate span.phpdebugbar-widgets-filename,
div.phpdebugbar div.phpdebugbar-widgets-sqlqueries li.phpdebugbar-widgets-list-item.phpdebugbar-widgets-sql-duplicate span.phpdebugbar-widgets-database,
div.phpdebugbar div.phpdebugbar-widgets-sqlqueries li.phpdebugbar-widgets-list-item.phpdebugbar-widgets-sql-duplicate span.phpdebugbar-widgets-duration,
div.phpdebugbar div.phpdebugbar-widgets-sqlqueries li.phpdebugbar-widgets-list-item.phpdebugbar-widgets-sql-duplicate span.phpdebugbar-widgets-memory,
div.phpdebugbar div.phpdebugbar-widgets-sqlqueries li.phpdebugbar-widgets-list-item.phpdebugbar-widgets-sql-duplicate span.phpdebugbar-widgets-row-count,
div.phpdebugbar div.phpdebugbar-widgets-sqlqueries li.phpdebugbar-widgets-list-item.phpdebugbar-widgets-sql-duplicate span.phpdebugbar-widgets-copy-clipboard,
div.phpdebugbar div.phpdebugbar-widgets-sqlqueries li.phpdebugbar-widgets-list-item.phpdebugbar-widgets-sql-duplicate span.phpdebugbar-widgets-stmt-id {
    color: var(--color-gray-500);
}
```

sebdesign:
I think `--color-gray-600` is nicer for the background. That brownish/poop doesn't look nice with the Dracula theme or Laravel colors.

sebdesign:
I'd prefer `--color-gray-600` but the yellow doesn't look that good against Dracula theme or Laravel colors.

PR #1085, opened 2020-08-16
Fix routes where $this->app doesn't exist

fletch3555:
Resolving outstanding question from #1055

If routes file is used (i.e. `include()`/`require()`) where `$this->app` doesn't exist, it would break.  This fixes that.


fletch3555:
Thanks!

barryvdh:
So this version does work now?

fletch3555:
Yes it does, thank you.  I planned to commit this one prior to #1055 being merged.  This is what I had tested locally, so it definitely works

PR #1088, opened 2020-08-16
Test injecting Debugbar

barryvdh:
Test some actual HTTP calls to see if it's injected.


barryvdh:
I'm guessing either the environment or running in console is not correctly overridden. Alternatively we could override isEnabled directly.

PR #1102, opened 2020-09-07
Fix "array to string conversion" in 3.5.0

LukeTowers:
This commit broke the debugbar: https://github.com/barryvdh/laravel-debugbar/commit/51c8ea3ab2c56664ebc9718ca68e342827bd7851#diff-8a2d53e88e43b04de0cc4f64882d91feR63 since sprintf() uses multiple arguments, it doesn't accept an array of arguments.

@barryvdh if you could review this ASAP that would be great


barryvdh:
Weird, I only automatically reformatted to psr-12 :/

PR #1114, opened 2020-10-02
Add debug_backtrace_limit configuration option

pataar:
This adds a configuration option for the debug_backtrace_limit. In some cases 50 is too low.


pataar:
@barryvdh What do you think of this addition?

pataar:
Thanks for the merge!

PR #1119, opened 2020-10-06
Provide an option to hide built-in routes from the list command

### LINKED ISSUE #1046: Unnecessary routing information is displayed by "artisan route:list" command.
I'm always helped by a great tool Laravel Debugbar.

But, I ran into a bit problem.
When I use the "php artisan route:list" command to display the routing information, I get the "_debugbar" information that I don't need.

The content is here.
laravel-debug

The routing information starting with "_debugbar" is I don't need it.
Is it possible to hide this information?

By the way, I tried to set  "APP_DEBUG=false" in the .env file and disabled the Laravel Debugbar, but it was still displayed.

<ENV>
Laravel Framework 6.18.13
barryvdh/laravel-debugbar - version: v3.3.3

miken32:
This resolves #1046 by overriding the built-in `artisan route:list` command to filter out the `_debugbar` routes. This functionality is enabled by a config flag, which is false by default.


miken32:
Seems to fail some tests with Lumen. I don’t have any experience with that, and the tests aren’t giving any backtracks to see where the problem might be.

sandervankasteel:
@barryvdh would an accept this PR, if I were to PR was complete with Lumen support ?

The reason I am asking, is because I made a PR towards @miken32 branch (which this PR uses) fixing the broken Lumen testing and I would love to get this merged :smiley:

barryvdh:
I'm sorta okay with this, but is this the proper way to extend it? Seems like there can be only one package to do this.

sandervankasteel:
> I'm sorta okay with this, but is this the proper way to extend it? Seems like there can be only one package to do this.

Now that you mention it, I am unsure if this is the "proper way" to do it. I kinda assumed the original PR creator took this into consideration.

But since this PR is still in a state of "limbo", I will investigate this and get back to you on this (and if needed modify the code).

barryvdh:
Gonna hold off on this, doesn't feel totally right like this.

PR #1120, opened 2020-10-08
Use of the same methods as in the whole class

bu4ak:
image


barryvdh:
The data was already encoded as json. This double-encoded it. Reverted an will tag a new version.

PR #1138, opened 2020-12-18
Remove unnecessary, deprecated defer codes

tanmuhittin:
Since Laravel 5.8 , deferred providers should implement DeferrableProvider contract instead of using $defer . $defer is deprecated.
https://laravel.com/docs/5.8/upgrade#deferred-service-providers

Also since this provider is not deferred It is unnecessary to have a provides method
Ref : https://laravel.com/docs/8.x/providers#deferred-providers


tanmuhittin:
Looks like the test failure is irrelevant with this PR. Could you please check.

PR #1141, opened 2021-01-05
Add AppStorage driver

mpociot:
This PR adds an additional "AppStorage" driver, which allows you to connect to the open-source Laravel DebugBar [companion app](https://github.com/beyondcode/laravel-debugbar-companion)

Screenshot 2021-01-05 at 12 18 16

The companion app works by opening a socket on the computer running the application. The Laravel DebugBar then sends the collected data JSON encoded to this socket, where the app listens for the data and displays it.

Since finding, clearing, or getting stored data does not apply in this case, I left those methods empty. The app/UI takes care of this.

The socket sending implementation itself is taken from the Symfony Dump Server, which works in a similar way.

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -34,12 +34,12 @@ return [
      */
     'storage' => [
         'enabled'    => true,
-        'driver'     => 'file', // redis, file, pdo, app, custom
+        'driver'     => 'file', // redis, file, pdo, socket, custom
         'path'       => storage_path('debugbar'), // For file driver
         'connection' => null,   // Leave null for default connection (Redis/PDO)
         'provider'   => '', // Instance of StorageInterface for custom driver
-        'hostname'   => '127.0.0.1', // Hostname to use with the "app" driver
-        'port'       => 2304, // Port to use with the "app" driver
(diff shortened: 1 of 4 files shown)


barryvdh:
Thanks! Would SocketDriver be more appropriate? 'app' seems a bit vague for me.

Is it Laravel specific? Otherwise we could merge the driver itself in the base debugbar package.

mpociot:
It would also work in the generic debugbar package, but there should still be a Laravel specific driver, that can provide additional metadata - like the application name and base path.
I use this information to group incoming requests and to allow users to open stack trace links in their favorite editor.
But this metadata could also move to the `meta` attribute in this package.
And SocketDriver is fine as well!

mpociot:
The debugbar itself only shows the latest/current request. So no AJAX requests will be visible in the bar itself, but only in the companion app.
I changed the name of the driver / config setting 👍

barryvdh:
Okay, I might think about adding a Chain driver that loops through the file + socket driver to store/retrieve the data, but I guess this will do for now :)

barryvdh:
v3.5.2 is tagged with this included :)

PR #1152, opened 2021-02-11
When config inject is false, not run store

gpibarra:
I want save all request (html) but i dont want inject bar


barryvdh:
Hmm I think this was intended.

Issue #1107, opened 2020-09-20
After debugbar installition middleware broken


burakucan:
after installing debugbar route:list commend write whole middleware thing line by line instead of write a just web

PR #1155, opened 2021-02-22
Fix route:list middleware full classes

gruz:
https://github.com/barryvdh/laravel-debugbar/issues/1107

Currently with installed debug bar `route:list` expands  middleware aliases (at least `auth` one) to a full class name. This is because the debug bar runs Http\Kernel in CLI scope and adds middlewares.

зображення


barryvdh:
This doesn't add anything?

PR #1165, opened 2021-04-05
Laravel Octane Compatibility

amaelftah:
base on @themsaid [comment here](https://twitter.com/themsaid/status/1379111701849763840) i have used $app instance instead of $this->app

and in [this tweet reply ](https://twitter.com/themsaid/status/1379112950410182658) why we need this change


themsaid:
Same with these two lines: https://github.com/barryvdh/laravel-debugbar/blob/master/src/ServiceProvider.php#L43-L44

amaelftah:
@themsaid thanks a lot :+1:

barryvdh:
Thanks!

Issue #1174, opened 2021-04-28
Octane: Target class [config] does not exist.


captenmasin:
Debugbar version: 3.5.5
Laravel version: 8.40.0
Octane version: 0.4.0

I'm not too well-versed in how Octane is supposed to handle app and config methods, but it seems to be erroring when using `$this->app['config']` ?

<img width="1214" alt="Screenshot 2021-04-28 at 11 28 50 pm" src="https://user-images.githubusercontent.com/25454872/116480668-a8b6e600-a879-11eb-907e-7aea57969f66.png">

This is just a plain page with Debugbar and Octane enabled, nothing else funky going on

zhuchunshu:
I also encountered the same problem

barryvdh:
Can you test the dev version? https://github.com/barryvdh/laravel-debugbar/pull/1179

ajnorman-dev:
The dev version fixes the DB collector

ImJustToNy:
I also confirm that #1179 fixes the issue.

ImJustToNy:
I also confirm that thar PR fixes the problem.

PR #1179, opened 2021-05-12
fix: db collector octane compatibility

J0sh0nat0r:
This patch should address #1174, although it does so rather crudely. A better solution would maybe be to re-register the listeners every request? I'm interested to hear any feedback on this.


barryvdh:
I think the threshold config could be moved outside of the callback. Not really sure about the other call though.

PR #1190, opened 2021-06-01
Queries tab: show duration as a progressbar on a background

denis-chmel:
Hi guys, 

This change makes it easy to notice long queries:

Screen Shot 2021-05-31 at 8 54 28 PM

Please let me know what I can improve,
Thanks!

PR #1196, opened 2021-06-10
Add option to toggle query background on/off

Aberdeener:
Hi,

I saw #1190 and see how it can be helpful, but for my uses I found it better disabled.
I am sure other people would be the same.
Let me know if I can improve anything


barryvdh:
I'm okay with that, but prefer an option like the rest of the config (setDurationBackground())

Aberdeener:
Done, thanks @barryvdh

Issue #1170, opened 2021-04-19
debugbar not showing up on translated routes?


rezaffm:
Hi,

I am not sure if thats a bug or more an issue on my side.

I have an international project with some routes that include a locale prefix such as "/de". If I for instance visit the root of the project, I see the debugbar. If I visit the root with "/de" I don't see it.

I am sure there is something I missed - or is that the expected behavior?

Routes are created dynamically  with a loop like....

```
<?php

foreach(app()->get('translatable.locales')->all() as $locale) {

    $prefix = '';
    $name = '';

    if ($locale !== config('app.default_locale')) {
        $prefix = $locale;
        $name = $locale . '.';
    }

    Route::prefix($prefix)->name($name)->group(function() use ($locale) {

        // Home (Root) route
        Route::get('/', 'Front\HomeController@index')->name('home.index');

...

}
```

rezaffm:
any explanation?

Issue #1149, opened 2021-01-29
Load routes in ServiceProvider


netsnatch:
https://github.com/barryvdh/laravel-debugbar/blob/cae0a8d1cb89b0f0522f65e60465e16d738e069b/src/ServiceProvider.php#L33

`loadRoutesFrom()` should be in `boot()` method instead of `register` (https://laravel.com/docs/8.x/packages#routes)

rust-7:
@netsnatch is this causing a "Target class [App\Http\Barryvdh\Debugbar\ServiceProvider] does not exist." error ? because thats what I'm getting.

netsnatch:
@rust-7 No, it does not cause an error.

In your case, the problem is in the namespace - **App\Http**\Barryvdh\Debugbar\ServiceProvider

rust-7:
> @rust-7 No, it does not cause an error.
> 
> In your case, the problem is in the namespace - **App\Http**\Barryvdh\Debugbar\ServiceProvider

Where do I have to do the modification ?

netsnatch:
Check config/app.php file. And remove App\Http part from class name.
But maybe it's not your case.

rust-7:
> Check config/app.php file. And remove App\Http part from class name.
> But maybe it's not your case.

Well, the `App\Http` is not mentioned in the documentation. It just says `\Barryvdh\Debugbar\ServiceProvider::class,` But I tried with and without, nothing to do...

netsnatch:
It was mentioned in error message. _Target class [App\Http\Barryvdh\Debugbar\ServiceProvider] does not exist_
So the problem is not in config/app.php

App\Http\Barryvdh looks strange.

Issue #1125, opened 2020-10-30
Routes should be registered in the ServiceProvider::boot method


xel1045:
Debugbar's routes seems to be wrongly registered since PR #1055 release in v3.5.0 on September 7th. Normally, as describe in Laravel's documentation (see https://laravel.com/docs/8.x/packages#routes), routes should be registered in the `ServiceProvier`'s `boot` method, but they are now configured in the `register` method.

Since the update, Debugbar doesn't show up and I can see a lot of theses errors:

```
Debugbar exception: Route [debugbar.openhandler] not defined. 
```

I'm currently testing in Laravel 7.29.2 and when I downgrade Debugbar to v3.4.2, everything is OK.

xel1045:
Yes, the problem is still present.

xel1045:
Related to issue #1149

PR #1197, opened 2021-06-10
Load routes and commands in the service provider's boot method

xel1045:
This PR move routes and commands loading to the `boot` method to avoid incompatibility with other plugins (ex: #1170, #1149, #1125). 

As described in [Laravel's documentation](https://laravel.com/docs/8.x/providers#the-register-method), routes and commands should be loaded in the `boot` method to avoid loading things that are not ready yet:

> As mentioned previously, within the register method, you should only bind things into the service container. You should never attempt to register any event listeners, routes, or any other piece of functionality within the register method. Otherwise, you may accidentally use a service that is provided by a service provider which has not loaded yet.

It seems a regression since PR #1055.


pgrenaud:
Thank you! 🙏

jnoordsij (approved):

PR #1224, opened 2021-09-01
Feature/ide links full path

macbookandrew:
Thanks for an awesome dev helper utility! This is a minor suggestion to improve DX for the Views panel.

Problem: the “link” icon to an IDE (VS Code specifically) contains a relative path to the project root, but VS Code tells me that a file does not exist. It seems to expect an absolute path on the filesystem.

Solution: provide an absolute path to the file instead of a relative path.

### DIFF
--- a/src/DataCollector/ViewCollector.php
+++ b/src/DataCollector/ViewCollector.php
@@ -89,7 +89,7 @@ class ViewCollector extends TwigCollector
         ];
 
         if ($this->getXdebugLink($path)) {
-            $template['xdebug_link'] = $this->getXdebugLink($path);
+            $template['xdebug_link'] = $this->getXdebugLink(realpath($view->getPath()));
         }
 
         $this->templates[] = $template;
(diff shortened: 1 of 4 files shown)


src/DataCollector/ViewCollector.php
```diff
@@ -89,7 +89,7 @@ public function addView(View $view)
         ];
 
         if ($this->getXdebugLink($path)) {
-            $template['xdebug_link'] = $this->getXdebugLink($path);
+            $template['xdebug_link'] = $this->getXdebugLink(realpath($view->getPath()));
```
barryvdh, line 92: This is an paranthese too much

barryvdh: Nvm

barryvdh:
The tests didn't run properly, skipped them with https://github.com/barryvdh/laravel-debugbar/pull/1256 for now. I think it relies on the specific format being available, so maybe set that in the setUp()?
Also trailing comma was not for older PHP versions.

PR #1235, opened 2021-09-30
Merge route middleware with controller ones

### LINKED ISSUE #1234: Middleware attached in Controller constructors are not reported
Middleware configured in `RouteServiceProvider.php` or attached to routes/route groups in the routes file (e.g. `web.php`) are reported by Debugbar in the Route tab.

However Middleware configured in Controller constructors are not.
```
<?php

namespace App\Http\Controllers;

class MwController extends Controller
{
    public function __construct()
    {
        $this->middleware('mwtest');
    }

    public function index()
    {
        return 'Viewing the ::index() method after mwtest middleware runs';
    }
}
```

Middleware connected this way still work and run, but are not reported by Debugbar.
Is there a constraint around the lifecycle that makes it difficult to report these? Or could it be changed without too much difficulty? 

I tested this with Debugbar 3.6.2 on Laravel 8.52.0 and Laravel 8.61.0

Katenkka:
Fixes #1234 (see docs, starting from Laravel 5.5: https://laravel.com/api/5.5/Illuminate/Routing/Route.html#method_gatherMiddleware)

### DIFF
--- a/src/DataCollector/RouteCollector.php
+++ b/src/DataCollector/RouteCollector.php
@@ -98,7 +98,7 @@ class RouteCollector extends DataCollector implements Renderable
     {
         return implode(', ', array_map(function ($middleware) {
             return $middleware instanceof Closure ? 'Closure' : $middleware;
-        }, $route->gatherMiddleware()));
+        },  $route->gatherMiddleware()));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


mike-healy:
Fantastic! 👊

PR #1240, opened 2021-10-05
fix: Ignoring Debugbar on No content routes

### LINKED ISSUE #1231: The Content-Length is changed when no content
I have some endpoints that use no content response. I this case I have a problem with Content-Length.

`
    protected function noContentResponse(): Response
    {
        return response('', 204);
    }
`

also, I have the middleware to count the content length:

`public function handle(Request $request, Closure $next)`
`    {`
       ` /** @var Response $response */`
`        $response = $next($request);`

`        return $response->header('Content-Length', (string)strlen((string)$response->getContent()));`
 `   }`

When I call my endpoint I have to see the **Content-Length: 0**
but instead, I see the huge value.

after the debugging, I found out that the LaravelDebugbar changes this value.

Could you please check this issue?
Thank you

Lootjs:
this pr closes #1231

### DIFF
--- a/src/Middleware/InjectDebugbar.php
+++ b/src/Middleware/InjectDebugbar.php
@@ -13,6 +13,8 @@ use Symfony\Component\Debug\Exception\FatalThrowableError;
 
 class InjectDebugbar
 {
+    public const HTTP_NO_CONTENT = 204;
+
     /**
      * The App container
      *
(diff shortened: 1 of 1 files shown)


Lootjs:
> Thanks, but I think this could better be placed here:
> 
> https://github.com/barryvdh/laravel-debugbar/blob/fc69fd6293e4e9d0f0181bb56219e9b3b3c4dd1f/src/LaravelDebugbar.php#L773-L786
> 
> In that case the data is still collected, but just not injected (for ajax requests, history etc)

Added data collection. If the solution is not suitable, I will probably create a new pr.

PR #1251, opened 2021-11-04
Return the result from the callback in `measure` helper

kemp:
This updates the helper to match the measure function in `src/LaravelDebugbar.php`.

I came across this while attempting to use the helper function to measure the time it takes to receive data from an API.

With this change, the following function will now work (where it doesn't without):

```php
$data = measure('Api request', function() {  
    return Http::get("https://api.example.com/api/stats");
});                                                                       
```


kemp:
Thanks!

PR #1257, opened 2021-11-12
set xdebug ini

macbookandrew:
Context: https://github.com/barryvdh/laravel-debugbar/pull/1224#issuecomment-966907943, #1256

> The tests didn't run properly, skipped them with #1256 for now. I think it relies on the specific format being available, so maybe set that in the setUp()?

@barryvdh I tried using `ini_set('xdebug.file_link_format', 'vscode://file/%f:%l')` in `setUp()` and it just wouldn’t set it. I also tried the `setUpBeforeClass()` method from phpunit to hook in a bit earlier and that didn’t work either.

Setting it in the phpunit.xml.dist file was the only way I was able to add that in so it would run.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -28,5 +28,6 @@
     <php>
         <env name="DEBUGBAR_ENABLED" value="true"/>
         <env name="DB_CONNECTION" value="testing"/>
+        <ini name="xdebug.file_link_format" value="vscode://file/%f:%l"/>
     </php>
 </phpunit>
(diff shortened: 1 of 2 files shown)


barryvdh:
So do we still need to skip the test?

macbookandrew:
Oh, good point…I just removed.

barryvdh:
Yeah doesn't work :(

macbookandrew:
I’ve been using phpunit 9.5.x but looks like your tests are using phpunit 8.5.x.

The phpunit.dist.xml change does **not** work for phpunit 8.5 but php’s `ini_set()` in `setUp()` **does**. 🤷🏻‍♂️

macbookandrew:
Apparently that doesn’t work either.

Should I add the conditional skip-test back?

barryvdh:
Hmm I guess

PR #1271, opened 2021-12-21
Update Dependencies for Laravel 9

### LINKED ISSUE #1270: [Feature request] Add Support for Laravel 9
Laravel 9 is around the corner (due to launch early next year), so some people like me are already actively developing on the new Laravel 9 platform, whether to push out their updates to Laravel 9 as soon as possible or to avoid having to update a new project after only a few weeks when Laravel 9 finally comes around.

It would be great if the laravel-debugbar package supported it already.

arminlinzbauer:
Resolves #1270

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,11 +12,11 @@
     "require": {
         "php": ">=7.2",
         "maximebf/debugbar": "^1.17.2",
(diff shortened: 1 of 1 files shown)


barryvdh:
Thanks!

PR #1252, opened 2021-11-10
Make service provider deferrable to allow for custom (filesystem) service provider

PrinsFrank:
As described on a [similar PR in the ide-helper](https://github.com/barryvdh/laravel-ide-helper/pull/914), when a serviceprovider is not deferrable, the service provider of a package is registered before any custom service providers in a project are registered.

I am currently working on a project with a custom FileSystemServiceProvider, where requiring the debugbar causes an error in Container.php on line 879 `Target class [files] does not exist` and Container.php on line 877 `Class "files" does not exist`. This issue was also mentioned in #1112.

<details>
    <summary>Simplest bug reproduction in code</summary>

## ./app/Providers/CustomFileSystemProvider.php

```php
<?php
declare(strict_types=1);

namespace App\Providers;

use Illuminate\Filesystem\FilesystemServiceProvider;

class CustomFileSystemServiceProvider extends FilesystemServiceProvider
{
    // Overwrite anything in here, or keep this empty for a simple reproduction
}
```

## ./config/app.php

```php
'providers' => [
    //Illuminate\Filesystem\FilesystemServiceProvider::class,
    App\Providers\CustomFileSystemServiceProvider::class,
```
</details>

This PR adds the DeferrableProvider interface to the package's service provider, and adds the provided services to a `provides` method so these services can be registered when needed.

PR #1306, opened 2022-03-16
Move registering routes to boot method of ServiceProvider

jnoordsij:
According to [Laravel docs](https://laravel.com/docs/9.x/providers#writing-service-providers), registering routes should never happen within the `register` method, as only binding should happen there. In practice, this causes a problem when trying to override/extend the base Laravel `FilesystemServiceProvider`, which combined with registering this package triggers an `Illuminate\Contracts\Container\BindingResolutionException` with message `Target class [files] does not exist.`.

Therefore, moving the registering of routes to the `boot` method looks like a more robust approach to me. See also the example from the [Laravel docs on the `loadRoutesFrom` method](https://laravel.com/docs/9.x/packages#routes).

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -26,8 +26,6 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
         $configPath = __DIR__ . '/../config/debugbar.php';
         $this->mergeConfigFrom($configPath, 'debugbar');
 
-        $this->loadRoutesFrom(realpath(__DIR__ . '/debugbar-routes.php'));
-
         $this->app->alias(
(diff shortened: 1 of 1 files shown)


PrinsFrank:
This issue is also solved in #1252 by marking the serviceProvider deferrable.

jnoordsij:
@barryvdh any chance you have some time to check this? It should be a quick and clear fix I think.

jnoordsij:
Actually, I see https://github.com/barryvdh/laravel-debugbar/pull/1197 already proposed this earlier. Closing this in favor of that, hoping it can be merged soon!

Issue #1319, opened 2022-06-07
Package symfony/debug is abandoned, you should avoid using it. Use symfony/error-handler instead.


muriloelias:
Hello guys, is it possible to update this dependency without much impact?

Package symfony/debug is abandoned, you should avoid using it. Use symfony/error-handler instead.

Thanks

muriloelias:
Very good, Thanks Jubeki. 

I hope the new release will be released soon. Big hug.

yeroon:
Would love to see this merged as well!

leonnicklas:
When will be the next release?

Thx for merging, but still waiting for the official release including the commit #3372ed6

barryvdh:
So is the package causing issues? Or is it just that message?

leonnicklas:
Just the message but with the next release including the latest commit this should go away, right?

PR #1321, opened 2022-06-09
Remove symfony/debug dependency

Jubeki:
Solves #1319 

The Class `Symfony\Component\Debug\Exception\FatalThrowableError` no longer exists, therefor I just removed it and directly throw the Error instead of wrapping it in another Exception.

From Laravel 7 onwards the report and render function also support `Throwable` instead of only `Exception`

This may also be a point where support for Laravel 7 can be dropped. (If you want only PHP ^8.0 Support, than it could also be considered to drop Laravel 8)
Laravel 6 gets security fixes until September 6th, 2022
Laravel 8 gets bugfixes until July 26th, 2022, security fixes until January 24th, 2023
See https://laravel.com/docs/9.x/releases#support-policy

How you can test these changes:
```json
{
    "repositories": [
        {
            "type": "git",
            "url": "https://github.com/Jubeki/laravel-debugbar"
        }
    ],
    "require-dev": {
        "barryvdh/laravel-debugbar": "dev-migrate-to-symfony-error-handler"
    },
    "minimum-stability": "dev",
    "prefer-stable": true
}
```

### DIFF
--- a/src/Middleware/InjectDebugbar.php
+++ b/src/Middleware/InjectDebugbar.php
@@ -2,14 +2,13 @@
 
 namespace Barryvdh\Debugbar\Middleware;
 
-use Error;
 use Closure;
 use Exception;
 use Illuminate\Http\Request;
 use Barryvdh\Debugbar\LaravelDebugbar;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Debug\ExceptionHandler;
-use Symfony\Component\Debug\Exception\FatalThrowableError;
+use Throwable;
 
 class InjectDebugbar
 {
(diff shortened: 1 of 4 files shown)


src/Middleware/InjectDebugbar.php
```diff
@@ -84,11 +80,11 @@ public function handle($request, Closure $next)
      * (Copy from Illuminate\Routing\Pipeline by Taylor Otwell)
      *
      * @param $passable
-     * @param  Exception $e
+     * @param  Throwable $e
      * @return mixed
      * @throws Exception
      */
-    protected function handleException($passable, Exception $e)
+    protected function handleException($passable, $e)
```
rodrigopedra, line 87: Couldn't `$e` be type-hinted as `Throwable`?

Jubeki: Yeah you are right about that.

```suggestion
    protected function handleException($passable, Throwable $e)
```

AratKruglik:
Any chances to have this fix soon? )

barryvdh:
Is the symfony error handler even required then?

PR #1345, opened 2022-09-29
Shows the real path of the views in the QueryCollector

SudoGetBeer:
Hey,

really loving this package. 
Recently I tried to debug a n+1 query but it was triggered in a view. But the debugbar only showed the path of the compiled version.
It uses the PATH ENDPATH logic added to the end of the compiled view. That was added in 5.8. ( https://github.com/laravel/framework/pull/28117 ).

The regex logic is the same Nuno  used in this PR. https://github.com/laravel/framework/pull/44347


barryvdh:
I think it is supposed to show the original path already. I'm okay with the endpath, but for other compilers this would not work.

Issue #1354, opened 2022-10-18
"Views" tab not finding rendered views


browner12:
With the release of Laravel v9.36.0, the "View" tab is no longer showing any rendered views.

Haven't nailed down *why* yet, but most likely due to [PR 44487](https://github.com/laravel/framework/pull/44487).

![Screen Shot 2022-10-18 at 11 10 38 AM](https://user-images.githubusercontent.com/5232313/196485092-cfd711e4-42da-46a3-a4a9-8a9b825a8c4e.png)

nunomaduro:
Fixed on Laravel `v9.36.3`.

PR #1355, opened 2022-10-18
Ensures a simple `composer` is registered so view events get fired

nunomaduro:
This pull request starts to fix https://github.com/barryvdh/laravel-debugbar/issues/1354, by registering a simple `composer` so view events get fired when using Debugbar.

@browner12 Can you try this, and let me know how it goes?


browner12:
confirmed the views are showing again in the debugbar

nunomaduro:
Fixed on Laravel `v9.36.3`.

PR #1365, opened 2022-11-29
Support for Read-Only PDO in QueryCollector

iamacarpet:
Hello,

We've run into a few problems using Debugbar on our project as it is configured read-only.

To ensure there are no development mistakes, the read connection is configured with a live DB instance and the write connection is configured with invalid details, so it'll throw an error if anyone tries to write to the DB.

Debugbar appears to be trying to use the write connection rather than read, with read being Laravel's default for SELECT statements (and it would also be fine for EXPLAIN SELECT).

Would you consider replicating Laravel's default behaviour with this change to use the read PDO when nothing other than SELECTs are being used?

Regards,
iamacarpet

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -133,7 +133,7 @@ class QueryCollector extends PDOCollector
 
         $pdo = null;
         try {
-            $pdo = $connection->getPdo();
+            $pdo = preg_match('/^\s*(SELECT) /i', $query) ? $connection->getReadPdo() : $connection->getPdo();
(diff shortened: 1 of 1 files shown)


cheeeasy2501:
@barryvdh Same issue. Could you check this mr?

barryvdh:
I'm not sure what is happening. What is it reading?

Also, I don't think that bit is doing anything anymore. Cleaned it up in https://github.com/barryvdh/laravel-debugbar/pull/1844

The other instance shouldn't trigger in new laravel versions either.

PR #1366, opened 2022-11-29
QueryCollector: Exception -> Throwable

iamacarpet:
Hello,

For some reason, on PHP 8.1, PDOException isn't being caught as an `\Exception`.

Swapping to `\Throwable` properly catches the error so it is silent.

`Throwable` is supported in PHP 7, so it shouldn't cause a conflict with your minimum version requirement.

Regards,
iamacarpet

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -134,7 +134,7 @@ class QueryCollector extends PDOCollector
         $pdo = null;
         try {
             $pdo = $connection->getPdo();
-        } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


iamacarpet:
I can't really explain this, as `PDOException` is an extension of `RuntimeException`, itself an extension of `Exception`.

Screenshot 2022-11-29 11 49 42 AM

Screenshot 2022-11-29 11 48 43 AM

It appears to be non-fatal in any case, but when catching `\Exception`, the error appears in the logs, which isn't the case when catching `\Throwable`.

PR #1390, opened 2023-02-12
Refactor command

datlechin:
What changes?

- Change `Console` namespace to `Commands` namespace
- Use new artisan command component (images below)
- Remove unused import class in `ServiceProvider`
- Use arrow function at line 49 in `ServiceProvider` file

Before:

___

After:


jbidad (commented):
its better to locate command into the console(alike laravel structure)

PR #681, opened 2017-07-25
Show only duplicated queries

decadence:
- Added button for hide unique queries
- Change `background-color` for repeated queries to darker.

PR #1394, opened 2023-02-19
Show only duplicated queries

decadence:
This is resubmission of https://github.com/barryvdh/laravel-debugbar/pull/681 to quickly find only duplicated DB queries.
It seems difficult to me to find only duplicated queries every time. As we learned some other developers find this useful too.

I didn't change color of the duplicated query line this time because it maybe not desirable for others. Color will not matter if we have this button. It's visible only if there are duplicated queries.

If text requires too many space you or me can replace it with some font icon I guess.

**Off**
Screenshot_1
**On**
Screenshot_3


decadence:
Thanks!

Issue #1369, opened 2022-12-17
Non-string messages not rendering when exception raised


michprev:
- Laravel v9.44.0 (PHP v8.1.13)
- barryvdh/laravel-debugbar version - v3.7.0

## How to reproduce
Modify the [laravel example](https://laravel.com/docs/9.x#your-first-laravel-project) app by inserting the following lines anywhere into `resources/views/welcome.blade.php`:
```blade
@php
$arr = [
    'name' => 'John',
    'age' => 30,
    'cars' => [
        'Ford',
        'BMW',
        'Toyota'
    ]
];
\Debugbar::log($arr);

throw new Exception('Error Processing Request', 1);

@endphp
```

## Observed behavior
![image](https://user-images.githubusercontent.com/7281754/208244491-943fb6ff-952f-4102-873d-d132c76c053b.png)

## Expected behavior
With the `throw new Exception('Error Processing Request', 1);` line removed / commented out:
![image](https://user-images.githubusercontent.com/7281754/208244554-56e5420a-ce2a-4687-8efd-d7adb872ef30.png)

PR #1399, opened 2023-03-02
Fixed debug output if exception or php fatal error is thrown.

### LINKED ISSUE #1379: Debugbar showing but my object information not showing anything
Possibly a bug here.

When I am to display my object's info using Debugbar, its showing blank if the page has error. Means no data in case there's an error on page. See screenshots below:

My Code:
<code>
$str = "test string";
$obj = $CheckUrlExist;
\Debugbar::info($str); // This is a sting variable and it shows the string
\Debugbar::info($obj); // This an object. It won't show if there's any error
var_dump($obj); //var_dump is showing data for on the object, even with error. So there's no question of object being null
</code>

Pages where's there no error:
image
See, we have the string and the database model object.

Pages with errors on page is showing as:
image
See the yellow marked one is showing blank.  And it is because of the error underlined in red.

var_dump() method output for object, even though there's error on page.
image
var_dump details marked.

It would be beneficial if we can use Debugbar instead of var_dump in case of errors too.

Thank you

crysix:
This will fix the problem that debug output is hidden if an error or exception is thrown.
Fixes #1379 and #1369

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.idea
 /build
 /vendor
 composer.phar
--- a/src/Resources/laravel-debugbar.css
+++ b/src/Resources/laravel-debugbar.css
@@ -174,7 +174,7 @@ div.phpdebugbar-minimized {
 }
 
 div.phpdebugbar code, div.phpdebugbar pre, div.phpdebugbar samp {
-    background: none;
+    background: none !important;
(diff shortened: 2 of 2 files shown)


kiranemasters (approved):
I think its good to go as of now. Thanks.

info-emastersindia (approved):
Thanks

PR #1400, opened 2023-03-09
allow user to customize memory precision

browner12:
**This PR is dependent on if/when https://github.com/maximebf/php-debugbar/pull/530 gets merged.**

it would be nice to be able to see more precise memory usage, as some changes will not have a full MB effect.

this change keeps the current default of "0", but allows the user to customize it to their liking.

the code is a little verbose because the root package is not setup for method chaining in this class.


parallels999:
@barryvdh what about https://github.com/maximebf/php-debugbar/pull/575, it's an alternative to this

PR #1404, opened 2023-03-09
remove unused imports

browner12:


src/DataCollector/RouteCollector.php
```diff
@@ -5,8 +5,6 @@
 use Closure;
 use DebugBar\DataCollector\DataCollector;
 use DebugBar\DataCollector\Renderable;
-use Illuminate\Http\Request;
-use Illuminate\Routing\Route;
```
davpsh, line 9: Yes, it can be removed, since the import is not used anywhere, but I would suggest to keep the import of the Response class and use it here
https://github.com/barryvdh/laravel-debugbar/blob/4a0e5842456819caafad7e5195b2bfe880a95d6d/src/DataCollector/RouteCollector.php#L72
as Response::class instead of a string value

src/LaravelDebugbar.php
```diff
@@ -6,10 +6,8 @@
 use Barryvdh\Debugbar\DataCollector\CacheCollector;
 use Barryvdh\Debugbar\DataCollector\EventCollector;
 use Barryvdh\Debugbar\DataCollector\FilesCollector;
-use Barryvdh\Debugbar\DataCollector\GateCollector;
```
davpsh, line 9: The same suggestions as above - leave the import and use in the code instead of string values

src/DataCollector/RouteCollector.php
browner12: you're free to make a separate PR, but I'm going to leave this as is.

davpsh (approved):

browner12:
no interest apparently, closing

PR #1412, opened 2023-04-30
Add support for PHPStorm via IDE Remote Control plugin

raveren:
Jetbrains official plugin for linking code to IDE from a webpage:

https://plugins.jetbrains.com/plugin/19991-ide-remote-control

Since the `phpstorm://` protocol does not work for a large number of people (me included:)


barryvdh:
This should be in upstream debugbar already now. Thanks

PR #1419, opened 2023-05-23
Fix php 8.0 - must be a valid hashing algorithm

### LINKED ISSUE #1418: Php v 8.0 unsupported
I have updated dependencies and came across this exception:

hash(): Argument #1 ($algo) must be a valid hashing algorithm

Thrown from file: /DataCollector/QueryCollector.php:396
That line is: if (hash('xxh128', 'v2' . $path) == $hash || sha1('v2' . $path) == $hash)

Did a fast google search and it looks like xxh128 hash was added in php 8.1
https://php.watch/versions/8.1/xxHash

Same results are from tinker
image

I checked composer.json and it looks that php 8.0 should be supported. So i suppose this is a bug?

Thank you in advance for looking into this.

erikn69:
Closes #1418

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -392,8 +392,10 @@ class QueryCollector extends PDOCollector
             $this->reflection['viewfinderViews'] = $property;
         }
 
+        $xxh128Exists = in_array('xxh128', hash_algos());
+
         foreach ($property->getValue($finder) as $name => $path) {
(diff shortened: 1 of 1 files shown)


dav245 (approved):
Can confirm this does the trick.

parallels999 (approved):

angeljqv:
Hi @barryvdh, could you review this bug?? thanks

PR #1359, opened 2022-10-31
View Editor Link

ferdousanam:
Hello,
I have added editor link beside the view path which will open view file in editor.

image

I have also added `LaravelViewTemplatesWidget` by extending `PhpDebugBar.Widget`

I hope this will help others to quickly go to editor through debugbar.

Thanks.

PR #1427, opened 2023-07-14
Reverting #1359 unnecessary file

erikn69:
Reverting #1359 unnecessary file

It is better to use the default `widget.js`, in case of any update in that file we would be covered

`editorLink` is already supported
[DebugBar/Resources/widgets/templates/widget.js#L42-L44](https://github.com/maximebf/php-debugbar/blob/293d02c78fd74d97e740772b079c33d61b9e90a7/src/DebugBar/Resources/widgets/templates/widget.js#L42-L44)


erikn69:
hi @barryvdh, ready for review

Issue #1435, opened 2023-07-25
LivewireCollector isn't working due to Livewire v3 release.


akunbeben:
Context: I'm using Livewire v3 and Laravel debugbar is not working on the Livewire v3.
As we know that Livewire v3 is released, yes it is still in beta. But, soon we need to adjust that release.

I did little research about this:

Current LivewireCollector is listening to the `view:render` event from the (`\Livewire\Event` now it's `\Livewire\EventBus`) class. And the callback method is using the instance of `Illuminate\Contracts\View\View`. And then getting the instance
https://github.com/barryvdh/laravel-debugbar/blob/cf4554d1dd7805a6a737dcb0f620c80d36850108/src/DataCollector/LivewireCollector.php#L26-L28

The latest event from Livewire v3 is `render` instead of `view:render`, and it's returning an instance of `\Livewire\Component` not `Illuminate\Contracts\View\View`.

Then, the `$component->id` is now updated to `private $__id;`. Now we can `$component->getId()` instead of `$component->id`

I can now capture the Livewire v3 on my debugbar locally, **I'm planning to open a PR to achieve this** and keep it backward compatible with the previous Livewire version. I created this issue just to let anyone using Laravel debugbar and Livewire v3 know why this does not capture the Livewire on the Laravel Debugbar.

Feedback would be appreciated. Thanks!

muffycompo:
Yes you are spot on with the research, it is just a matter hearing what @barryvdh thinks of the changes and if there is a need to send a PR or it is something the team will like to add themselves.

Thanks again for catching that quick as I myself was caught on the WHY it was not working as intended.

barryvdh:
Sounds good!

muffycompo:
@akunbeben can you send the pull request you mentioned regarding the issue or you need some time to write it? I can offer to review it if you need a second set of eyes. Thanks again.

akunbeben:
Thanks @barryvdh @muffycompo! Here is the PR #1436✨

PR #1436, opened 2023-07-26
feat: update LivewireCollector to support Livewire v3

akunbeben:
Based on my explanation here #1435. This is to support Livewire v3 release.

To keep it backward compatible with the previous version of Livewire, I keep the old listener. 

Here I created a new listener based on the old listener: 
I removed the $data['view'] due to in the Livewire 3 is now we don't really need to have a view.
Then, `$component->getPublicPropertiesDefinedBySubClass()` is now moved to the `Livewire\Concerns\InteractsWithProperties::all`. So, `$component->all()`

I updated the `Barryvdh\Debugbar\LaravelDebugbar::isJsonRequest` to only check the header key instead of its value. Because the Livewire v3 `x-livewire` header is an empty string `""`, as you know it could be a falsy value

### DIFF
--- a/src/DataCollector/LivewireCollector.php
+++ b/src/DataCollector/LivewireCollector.php
@@ -12,6 +12,7 @@ use Illuminate\Support\Arr;
 use Illuminate\Support\Fluent;
 use Illuminate\Support\Str;
 use Livewire\Livewire;
+use Livewire\Component;
 
 /**
  * Collector for Models.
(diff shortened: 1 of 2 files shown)


akunbeben:
Thanks, @barryvdh! Do you mind it to share when it will be released?

barryvdh:
So the listen doesn't cause problems with livewire v2?

akunbeben:
Manually tested, yes it does not.

PR #1438, opened 2023-07-26
Livewire v3 Support

lrljoe:
Due to the removal of the $component->id function, which is replaced with $component->getId(), the current version of Debugbar returns a blank Livewire slot.

This PR matches against the "id" property if it exists, otherwise it uses the getId() to provide some backwards compatibility.

No tests created.

### DIFF
--- a/src/DataCollector/LivewireCollector.php
+++ b/src/DataCollector/LivewireCollector.php
@@ -27,16 +27,15 @@ class LivewireCollector extends DataCollector implements DataCollectorInterface,
         Livewire::listen('view:render', function (View $view) use ($request) {
             /** @var \Livewire\Component $component */
             $component = $view->getData()['_instance'];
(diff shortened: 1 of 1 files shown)


akunbeben:
I think you don't need to check this because the event name is completely different between v2 (`view:render`) and v3 (`render`).

So, if we use Livewire v2 it will not go through the `Livewire::listen('render')`
Same for the v3 will not go through the `Livewire::listen('view:render')`.

I already tested it manually both v2 and v3.

lrljoe:
Ah fair enough!  I've updated the PR accordingly, thanks for the pointer!

akunbeben:
The latest update is not published to Packagist yet, so we just need to wait patiently. 

For reference, you can read here #1435

parallels999:
This open PR has no changes

Issue #1452, opened 2023-09-07
DEBUGBAR can be ABUSED to obtain USER Credentials!


whipsterCZ:
**WARNING: using debugbar at TEST or PRODUCTION environment will lead to SECURITY BREACH...**

im using (i was using) DEBUGBAR at several pages.
But we were recently hacked - Debugbar has log of all requests INCLUDING LOGIN REQUESTS with PASSWORDS!!!!

Can you share this as ATTENTION 
Or at least add some warning to README.MD , i have not clue that it can be ABUSED like this.. 

DETAILS: 
attacker found debugbar enabled on our TEST environment, but we are using same credentials on PROD
so attacker had access to our ADMIN account .

```
(/_debugbar/open?method=POST&uri=*%2Flogin&max=20&offset=0)

I found this:

"password" => {plaintext-password!}
"email" =>  {adminEmail}
```

parallels999:
- https://github.com/barryvdh/laravel-debugbar/pull/1442
- https://github.com/barryvdh/laravel-debugbar/pull/1443

parallels999:
> Debugbar has log of all requests INCLUDING LOGIN REQUESTS

Disable debugbar on login, just enable it whe you need, i did add a debugbar enabler button

barryvdh:
This should already be fixed in the latest version. Browsing is disabled by default.
Note that you should still not show it on public sites, which was already mentioned in the config.

PR #1453, opened 2023-09-07
Hide sensitive information by config

erikn69:
#1452 shows a security problem on debugbar files, this PR offer posibility of hide data by config, sometimes can be used not default keys like `pass`, `token`, or others
Only `request_request.password` is hidden from `request_request`
https://github.com/barryvdh/laravel-debugbar/blob/27b088aaad2adba74a7a73a47cb8d11277b551e3/src/DataCollector/RequestCollector.php#L128-L130

**CC:** @barryvdh 

**UPDATE:** Now it hide keys on clockwork too, also it hide `authorization` token(security fix)

### DIFF
--- a/src/DataCollector/SessionCollector.php
+++ b/src/DataCollector/SessionCollector.php
@@ -5,20 +5,25 @@ namespace Barryvdh\Debugbar\DataCollector;
 use DebugBar\DataCollector\DataCollector;
 use DebugBar\DataCollector\DataCollectorInterface;
 use DebugBar\DataCollector\Renderable;
+use Illuminate\Support\Arr;
 
 class SessionCollector extends DataCollector implements DataCollectorInterface, Renderable
 {
     /** @var  \Symfony\Component\HttpFoundation\Session\SessionInterface|\Illuminate\Contracts\Session\Session $session */
     protected $session;
+    /** @var array */
+    protected $hiddens;
 
     /**
      * Create a new SessionCollector
      *
      * @param \Symfony\Component\HttpFoundation\Session\SessionInterface|\Illuminate\Contracts\Session\Session $session
+     * @param array $hiddens
      */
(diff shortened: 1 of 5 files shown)


oprudkyi:
Hi @barryvdh 
could you please take a look at this PR
it would be great to have configurable list of hidden keys in some scenarios (multiple qa, non-standard keys)

oprudkyi:
Hi @barryvdh

is anything can be done here to move it forward ?

src/DataCollector/SessionCollector.php
```diff
@@ -30,6 +35,13 @@ public function collect()
         foreach ($this->session->all() as $key => $value) {
             $data[$key] = is_string($value) ? $value : $this->formatVar($value);
         }
+
```
oprudkyi, line 38: Hi @erikn69 
this won't handle deep array values as they already converted to string, probably all collect() should look like 
```
    /**
     * {@inheritdoc}
     */
    public function collect()
    {
        $data = $this->session->all();

        foreach ($this->hiddens as $key) {
            if (Arr::has($data, $key)) {
                Arr::set($data, $key, '******');
            }
        }

        foreach ($data as $key => $value) {
            $data[$key] = is_string($value) ? $value : $this->formatVar($value);
        }


        return $data;
    }

```
i.e. apply hiddens before converting values to strings

oprudkyi:
Hi @erikn69 
also, probably worth to add the same support into src/Support/Clockwork/ClockworkCollector.php 
it includes the same data - cookies, headers etc

erikn69:
> also, probably worth to add the same support into src/Support/Clockwork/ClockworkCollector.php

@oprudkyi you are right, but I don't know if this PR is going to be merged, anyway I'm going to try

barryvdh (approved):

PR #1460, opened 2023-09-11
Support class resolver for open storage

parallels999:
For config cache, 
Wouldn't it be better to be able to use a resolver by class name instead of a callback method?
So, on `storage.open` we can set a string class name, where is a `resolve` method
Or in `DEBUGBAR_OPEN_STORAGE ` we just can set the fqn of the class

### DIFF
--- a/src/Controllers/OpenHandlerController.php
+++ b/src/Controllers/OpenHandlerController.php
@@ -24,6 +24,10 @@ class OpenHandlerController extends BaseController
             return call_user_func($open, [$request]);
         }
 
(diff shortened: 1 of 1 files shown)


barryvdh:
Can't this be done in the callback? This doesn't seem like a convention of Laravel?

parallels999:
>Can't this be done in the callback? 

https://github.com/barryvdh/laravel-debugbar/blob/7746529baa332e2edfaf6d5a69841e421f57553d/src/Controllers/OpenHandlerController.php#L23-L25
yes, but the problem is that if you set a closure on config, then you can't cache that config, because closure is not serializable

PR #1462, opened 2023-09-14
Show queries memory usage

erikn69:
Adds memory usage  on queries tab
image

image

the js widget already supports this


jose123v:
why `memory_usage` flag?
I think it should be a default functionality.
@barryvdh thoughts??

barryvdh:
I think real usage indeed.

barryvdh (approved):

barryvdh:
Nevermind, false show the used memory instead of allocated.

PR #1467, opened 2023-09-19
Set memory baseline on MemoryCollector

### LINKED ISSUE #1434: debug bar showing 19-21mb memory usage on clean install laravel (welcome page)
I have tested it on laravel 8 , 9 and 10. all of them showing 19-21mb usage on welcome page while having fresh installation.
Is this normal or something is wrong with package. I have php v8.1.17 installed. i did composer update as well. but issue remains same. I have watched few installation tutorials. They have less memory usage than mine for same laravel version.

erikn69:
Closes #1434
Complements of https://github.com/maximebf/php-debugbar/pull/559

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -189,6 +189,7 @@ return [
 
     'options' => [
         'memory' => [
+            'reset_peak' => false,     // run memory_reset_peak_usage before collecting
             'with_baseline' => false,  // Set boot memory usage as memory peak baseline
         ],
         'auth' => [
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -188,6 +188,9 @@ return [
      */
 
     'options' => [
+        'memory' => [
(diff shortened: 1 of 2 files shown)


cesarreyes3:
hi @barryvdh, could you check this?, thanks

Rattone:
Why is this not being merged?

cesarreyes3:
@Rattone hi, maybe @barryvdh has no time to review it

barryvdh:
I'm not really sure on the use case here, but okay.

barryvdh (approved):

barryvdh:
Yeah I understand that, but I'm not really sure if this is the best way. It feels a bit weird that it's excluding the bootstrap. But for Octane etc I do understand why you would need to reset the usage.

But not really sure why the memory is so different. I'd rather just 'have it work' correctly without config, but not sure. If Linux or windows reservers more memory, it would be sufficient to use the 'real' usage instead of the allocated usage?

PR #1469, opened 2023-09-20
Support editor link on ConfigCollector

erikn69:
Add editor link on the configs tab
image


barryvdh:
I think this is slightly overkill, and also might not always be correct if the config is coming from other sources.

PR #1471, opened 2023-09-22
`route_middleware` from config

systemsolutionweb:
middleware to the debugbar routes from config


angeljqv (approved):

barryvdh:
But we would need to add a config then.

systemsolutionweb:
> But we would need to add a config then.

to document it, yes, so anyone could find it more easily if I needed it

PR #1473, opened 2023-09-29
(fix) Inject debugbar at the first `</head>`

### LINKED ISSUE #1429: Issue with adding of the debugbar at the last occurence in head
In my project I am showcasing some ways to do specific tasks in Magento 1 and 2.
So, in the php file:
https://github.com/barryvdh/laravel-debugbar/blob/master/src/LaravelDebugbar.php
At line: 914-915 I see the function is at follows:
// Try to put the js/css directly before the < /head>
$pos = strripos($content, '< /head>');

If I add in the content another < /head> then the entire pages does not work properly anymore.

What is the purpose of finding the last occurrence and not the first?

erikn69:
Closes #1429

Try adding this on any view: `<textarea></head></textarea>`,
then debugbar stop working

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -927,7 +927,7 @@ class LaravelDebugbar extends DebugBar
         $widget = $renderer->render();
 
         // Try to put the js/css directly before the </head>
-        $pos = strpos($content, '</head>');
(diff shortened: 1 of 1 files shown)


barryvdh:
Yeah I changed it to stripos for case insensitivity

PR #1474, opened 2023-10-04
Test against PHP 8.3

sergiy-petrov:


jnoordsij (approved):
As discussed in https://github.com/barryvdh/laravel-debugbar/pull/1476#discussion_r1357051542, there probably should be a follow-up to correct the implementation of the `__toString` method in the `LaravelDebugbar` class. Given that it is unrelated to this PR and this PR only changes the styling (through an automated fix), I think this is ready to merge!

barryvdh:
Thanks!

PR #1476, opened 2023-10-08
GitHub workflow improvements

jnoordsij:
(This is a follow-up to #1474, as I was doing this myself, then found out there already was a PHP 8.3 PR, so please merge that first; this is based on the branch to prevent merge conflicts).

When updating the workflow, I noticed a few things that could be slightly improved:
- use the latest stable PHP version for the `fix-cs` workflow
- remove the explicit `nesbon/carbon` version requirement in `run-tests` as it is no longer needed
- Add Lumen 10 to `run-integrations-test`, which was probably already meant to be there

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -348,7 +348,7 @@ class LaravelDebugbar extends DebugBar
                         if (!app(static::class)->shouldCollect('db', true)) {
                             return; // Issue 776 : We've turned off collecting after the listener was attached
                         }
-                        
+
                         $bindings = $query->bindings;
                         $time = $query->time;
                         $connection = $query->connection;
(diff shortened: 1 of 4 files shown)


src/LaravelDebugbar.php
```diff
@@ -470,15 +470,20 @@ public function __construct($transport, $laravelDebugbar)
...
                             );
                         }
-                        protected function doSend(SentMessage $message): void {}
-                        public function __toString(): string{ $this->originalTransport->__toString(); }
+                        protected function doSend(SentMessage $message): void
+                        {
+                        }
+                        public function __toString(): string
+                        {
+                            $this->originalTransport->__toString();
```
erikn69, line 485: return missing
```diff
- $this->originalTransport->__toString();
+ return $this->originalTransport->__toString();
```

jnoordsij: This should be addressed in #1474, where I based this on. I can rebase once that is done and merged.

Also this looks likes an automated fix, so might also need to be reported/fixed in the autofixer.

parallels999 (approved):

erikn69 (approved):

src/LaravelDebugbar.php
jnoordsij: Actually after a second look I noticed this already was incorrect before the auto-styling was applied. Probably might be better to fix this in another PR once #1474 and this have been merged.

PR #1494, opened 2023-11-20
Adds config option to disable the display of parameters in views.

nimmneun:
So I was playing with filamentphp last weekend and noticed some heavy browser lags for basically all list views with 100s of blade components (which is basically all xD).

After profiling I noticed a heavy lag when rendering the parameter list.
image
`vendor/maximebf/debugbar/src/DebugBar/Resources/widgets/templates/widget.js:50`

Easy way around this would be an option to disable params from this package.

I profiled this on 2 Laptops ... with the same result.

To reproduce and ensure I'm not the only one suffering from this, just return before params are written into the dom element:
`vendor/maximebf/debugbar/src/DebugBar/Resources/widgets/templates/widget.js:50`
image

... just figured imma rename this config to with_params, default true 🙈 one sec pls 

If I'm the only one experiencing this, please downvote this PR with a 👎 xD

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -213,6 +213,7 @@ return [
         'views' => [
             'timeline' => false,  // Add the views to the timeline (Experimental)
             'data' => false,    //Note: Can slow down the application, because the data can be quite large..
+            'exclude_params' => false, // Enable to reduce browser lag for views with many blade components
             'exclude_paths' => [], // Add the paths which you don't want to appear in the views
         ],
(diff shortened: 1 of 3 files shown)


barryvdh:
What do you think about https://github.com/barryvdh/laravel-debugbar/pull/1503 ?

nimmneun:
@barryvdh looking good + more flexible 👍

nimmneun:
solved by #1503

PR #1503, opened 2024-02-09
Reduce View/Templates impact on performance

barryvdh:
- By default, don't show parameters. Add new config to allow just the keys (previous default)
- By default, group the templates so you don't render 50 of the same lines
- Add filament to exclude because of the many components you probably don't need.

Before:
image

After:
image


barryvdh:
For the Twitter users, tweaked it a bit to use similar links to phpstorm etc evewhere. And only group after a threshold (50 by default).
image
image

PR #1258, opened 2021-11-12
Add support editor

VictoRD11:
Hi,

This pull request is a proposition to add a link to the editor for a given file and line, if available.

Thanks ;)

PR #1506, opened 2024-02-10
Set default base path for xdebug replacements

barryvdh:
I'm still not really sure if this is correct. It seems so be copied from Ignition in #1258 by @VictoRD11 ?
But should both default to the base_path?
Eg. in Laravel Herd this works fine by default, with `getRemoteServerReplacements()` looking like this:
image

But not sure if this is intended like this. And should we default to the Ignition .env?
/cc @ erikn69

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -85,8 +85,8 @@ return [
     |
     */
 
-    'remote_sites_path' => env('DEBUGBAR_REMOTE_SITES_PATH', ''),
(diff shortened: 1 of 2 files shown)


barryvdh:
Okay like this it seems to work without setting anything in local development (eg. Herd/Valet), and just setting the local settings for ignition fixes both ignition and debugbar the same way :)

barryvdh:
Yeah it uses the same logic anyways

PR #1526, opened 2024-02-14
Allow Debugbar to work with runtime connection

deleugpn:
### Issue

In a clean Laravel application, if you erase the `database.default` and the `database.connections` block, Laravel Debugbar will not be able to collect database information.

### Root Cause

Laravel Debugbar automatically injects the `InjectDebugbar` as a Global Middleware and adds Database Event Listeners through the `DatabaseManager $db` class which goes through `__call()` and tries to resolve the default connection, leading to an error.

The fact Debugbar goes through `__call()` makes it trigger a connection. 

### Proposal

Use the event listener directly to listen to relevant events without causing a connection to be established. If the database is never used, it will never connect and events will not be dispatched.

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -299,7 +299,11 @@ class LaravelDebugbar extends DebugBar
         }
 
         if ($this->shouldCollect('db', true) && isset($this->app['db'])) {
+            /** @var \Illuminate\Database\DatabaseManager $db */
             $db = $this->app['db'];
+            /** @var \Illuminate\Events\Dispatcher $events */
+            $events = $this->app['events'];
+
             if (
                 $debugbar->hasCollector('time') && $this->app['config']->get(
                     'debugbar.options.db.timeline',
@@ -346,7 +350,7 @@ class LaravelDebugbar extends DebugBar
             $this->addCollector($queryCollector);
 
             try {
-                $db->listen(
+                $events->listen(
                     function (\Illuminate\Database\Events\QueryExecuted $query) {
                         if (!app(static::class)->shouldCollect('db', true)) {
                             return; // Issue 776 : We've turned off collecting after the listener was attached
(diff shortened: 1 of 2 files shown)


src/LaravelDebugbar.php
```diff
@@ -299,7 +299,11 @@ function ($level, $message = null, $context = null) use ($logger) {
         }
 
         if ($this->shouldCollect('db', true) && isset($this->app['db'])) {
+            /** @var \Illuminate\Database\DatabaseManager $db */
```
erikn69, line 302: After this change, `$db` seems unused, please check it, Maybe it can be removed

erikn69:
In my opinion, there may be a good reason why they did it that way.
Maybe in certain scenarios there is no `$this->app['events']`. (`isset($this->app['events'])`)
https://github.com/barryvdh/laravel-debugbar/blob/0f74197af08c1d8945dec2d5103453e333b74af2/src/LaravelDebugbar.php#L222
https://github.com/barryvdh/laravel-debugbar/blob/0f74197af08c1d8945dec2d5103453e333b74af2/src/LaravelDebugbar.php#L232
https://github.com/barryvdh/laravel-debugbar/blob/0f74197af08c1d8945dec2d5103453e333b74af2/src/LaravelDebugbar.php#L529
https://github.com/barryvdh/laravel-debugbar/blob/0f74197af08c1d8945dec2d5103453e333b74af2/src/LaravelDebugbar.php#L447

deleugpn:
I added an `isset($this->app['events'])` check there to avoid potential breaks. If `$app['events']` is not set, then trying to listen to a database event will not work, regardless of the connection method, so this looks pretty safe, imo.

erikn69:
>If $app['events'] is not set, then trying to listen to a database event will not work, regardless of the connection method, so this looks pretty safe, imo.

Good point

PR #1527, opened 2024-02-14
Support Inertia ajax calls

erikn69:
After https://github.com/barryvdh/laravel-debugbar/pull/1525#issuecomment-1943878274, I realized that only first load uses a blade view, from that point all views are loaded through ajax, now ajax calls shows the inertia view
image
I have removed junk code that does nothing, I also fixed some merging errors, thanks to that, the editor link opens the inertia file instead of the blade layout (_this layout is the same for all inertia views_)

I see some features in this collector, I did try to keep everything in order, but check it

### DIFF
--- a/src/DataCollector/ViewCollector.php
+++ b/src/DataCollector/ViewCollector.php
@@ -92,16 +92,14 @@ class ViewCollector extends TwigCollector
             if (isset($data['props'], $data['component'])) {
                 $name = $data['component'];
                 $data = $data['props'];
-                $oldPath = $path;
 
-                if (!@file_exists($path = resource_path('js/Pages/' . $name . '.js'))) {
-                    if (!@file_exists($path = resource_path('js/Pages/' . $name . '.vue'))) {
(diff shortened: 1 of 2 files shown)


src/DataCollector/ViewCollector.php
```diff
@@ -58,35 +58,9 @@ public function getWidgets()
      */
     public function addView(View $view)
     {
-        $name = $view->getName();
-        $path = $view->getPath();
-        $data = $view->getData();
-        $shortPath = '';
-        $type = '';
+        list($name, $type, $data, $path) = $this->getInertiaView($view->getName(), $view->getData(), $view->getPath());
```
barryvdh, line 61: This title seems a bit odd if it's not for Inertia only?

erikn69: Yes, the method try inertia view, opposite case, fallback the same args
I'm using Google Translate and I didn't know what name to give it, plus I reused it from the ajax calls

barryvdh:
Tweaked it a bit

PR #1531, opened 2024-02-15
Add files back for BC compact with 3rd party extensions

barryvdh:


systemsolutionweb:
Jobs collector was introduced on 3.10, I don't see the need to "reintroduce" it, no 3rd party extension had time to add a dependency for it

systemsolutionweb:
ModelsCollector yes, it was since 3.5

PR #1534, opened 2024-02-15
Make opt-in events data collecting

erikn69:
There are pages that can have more than a thousand events, this option allows you to deactivate the capture of the parameters/listeners of each event if it is not necessary

image

Same as cacheCollector 
https://github.com/barryvdh/laravel-debugbar/blob/d1a48965f2b25a6cec2eea07d719b568a37c9a88/src/DataCollector/CacheCollector.php#L30

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -245,6 +245,9 @@ return [
         'symfony_request' => [
             'hiddens' => [], // hides sensitive values using array paths, example: request_request.password
         ],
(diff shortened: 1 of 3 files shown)


barryvdh:
Should we default to false?

erikn69:
> Should we default to false?

To avoid loading too much information, yes, it should
I left it true so as not to change existing functionality.
I'll change it then

PR #1535, opened 2024-02-15
Use Ignition defaults as fallback on config.php

PaolaRuby:
Easy setup if you already have Ignition configured
https://github.com/barryvdh/laravel-debugbar/blob/d1a48965f2b25a6cec2eea07d719b568a37c9a88/config/debugbar.php#L89
https://github.com/spatie/laravel-ignition/blob/6ad3d229c342d881e803b08b8540fb8d1c980b39/config/ignition.php#L37-L43
https://github.com/spatie/laravel-ignition/blob/6ad3d229c342d881e803b08b8540fb8d1c980b39/config/ignition.php#L52-L56

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -60,7 +60,7 @@ return [
     |
     */
 
-    'editor' => env('DEBUGBAR_EDITOR', 'phpstorm'),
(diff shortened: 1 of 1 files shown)


barryvdh:
I reverted the theme setting in https://github.com/barryvdh/laravel-debugbar/commit/d2c33f073e0387074b6c4bd24ac907b706fba452 because I don't think they should be the same usually.

PR #1536, opened 2024-02-16
More descriptive messages on GateCollector

erikn69:
Sin nombre

image

Depends on https://github.com/maximebf/php-debugbar/pull/601
It is not a breaking change, you do not need to modify the dependency version


systemsolutionweb:
Colors don't match with dark theme, IMO it isn't readable enough

erikn69:
image

How about using the Dracula Theme colors?

barryvdh:
That looks better

PR #1546, opened 2024-02-23
Use dark theme by default

barryvdh:
In my opinion, the dark theme is better, even without having the website in dark mode. Thoughts?

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -308,7 +308,7 @@ return [
      | Switches between light and dark theme. If set to auto it will respect system preferences
      | Possible values: auto, light, dark
      */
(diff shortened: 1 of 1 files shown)


erikn69:
For most dark themes are better, but it is configurable, 
and this is a debug-only package, it's not like it's for the public

cotiga:
I don't like dark mode, please give us the choice and leave light mode as default...
Personally, I'm tired of having to waste my time all the time with the whims of the devs :p

barryvdh:
You can set is with config or an env though?

barryvdh:
See #1594

PR #1548, opened 2024-02-24
PHP8.2 explode(): Passing null to parameter #2 ($string) of type string is deprecated

grasebit:
Hi.

1. deprecated by PHP8.2
    * `local.WARNING: explode(): Passing null to parameter #2 ($string) of type string is deprecated in /*/vendor/barryvdh/laravel-debugbar/src/LaravelDebugbar.php`
2. use PHP-CS-Fixer (LaravelPint) to config/debugbar.php

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -36,14 +36,14 @@ return [
      | Specify a callback if you want to limit based on IP or authentication.
      */
     'storage' => [
-        'enabled'    => true,
-        'open'       => env('DEBUGBAR_OPEN_STORAGE', false), // bool/callback.
-        'driver'     => 'file', // redis, file, pdo, socket, custom
-        'path'       => storage_path('debugbar'), // For file driver
-        'connection' => null,   // Leave null for default connection (Redis/PDO)
-        'provider'   => '', // Instance of StorageInterface for custom driver
-        'hostname'   => '127.0.0.1', // Hostname to use with the "socket" driver
-        'port'       => 2304, // Port to use with the "socket" driver
+        'enabled' => true,
+        'open' => env('DEBUGBAR_OPEN_STORAGE', false), // bool/callback.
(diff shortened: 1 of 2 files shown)


erikn69:
```php
$this->app['config']->get('debugbar.remote_sites_path', '')
``` 
if null get returns the empty string (`''` is a fallback default), so, it would never be null
How did you get that error message?

sebastiaanluca:
Can confirm the issue, even with an identical config file as in the repo.

```
// returns `null`
$this->app['config']->get('debugbar.remote_sites_path')

// returns `null`
$this->app['config']->get('debugbar.remote_sites_path', '')

// returns null 🤷‍♂️
$this->app['config']->get('debugbar.remote_sites_path', 'uhmmm')
```

It should fall back to the second parameter, but for some reason it doesn't and you get the deprecation warning in the log.

grasebit:
env() function returns the value if the key exists.


```
// config/debugbar.php
'remote_sites_path' => env('DEBUGBAR_REMOTE_SITES_PATH')
// remote_sites_path is null

// src/LaravelDebugbar.php
$foo = $this->app['config']->get('debugbar.remote_sites_path', '');
// $foo is null

$bar = $this->app['config']->get('debugbar.remote_sites_path') ?? '';
// $bar is empty string
```


```
// config/debugbar.php
'remote_sites_path' => env('DEBUGBAR_REMOTE_SITES_PATH', '')
// remote_sites_path is empty string

// src/LaravelDebugbar.php
$foo = $this->app['config']->get('debugbar.remote_sites_path', '');
// $foo is empty string

$bar = $this->app['config']->get('debugbar.remote_sites_path') ?? '';
// $bar is empty string
```

parallels999:
Maybe they would merge it if you removed the other unnecessary changes

PR #1552, opened 2024-03-01
Check if view path is set

barryvdh:


erikn69:
This is actually a merge error, the real code was
```php
if ($path && $this->getXdebugLinkTemplate()) {
    $template['xdebug_link'] = $this->getXdebugLink($path);
}
```
`getXdebugLink` already does `realpath`
https://github.com/barryvdh/laravel-debugbar/blob/1e7ddcc1aadc4ca3219a3cd42cadef6b0ab26d43/src/DataCollector/ViewCollector.php#L62

erikn69:
I did rebase #1527, there is the fix too

barryvdh:
Not sure if that is correct? The $path is different than the actual path, because $path is already normalized differently

erikn69:
In this case yes, there were many fusion errors, I corrected them in the other PR, #1527

PR #1569, opened 2024-03-11
Fix #1568 MailCollector Regression

### LINKED ISSUE #1568: DebugBarException "'mail' is not a registered collector" on 3.11.0
When sending mails within my Laravel application with debugbar active, the following `DebugBarException` is shown: `'mail' is not a registered collector`.

My assumption is this is a regression introduced in #1521, which assumes the collector added in the mail scope to be `'mail'`. However, the current name of this collector is `'symfonymailer_mails'` (see https://github.com/maximebf/php-debugbar/blob/a8694a4e0ee2291d13de5ff1d700d94393366511/src/DebugBar/Bridge/Symfony/SymfonyMailCollector.php#L66).

Possible solutions:
- rename at one of the two places
- partially revert changes from #1521

This issue is not present on 3.10.6.

erikn69:
Closes #1568

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -429,11 +429,11 @@ class LaravelDebugbar extends DebugBar
             try {
                 $this->addCollector(new SymfonyMailCollector());
                 $events->listen(function (MessageSent $event) {
-                    $this['mail']->addSymfonyMessage($event->sent->getSymfonySentMessage());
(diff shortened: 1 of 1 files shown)


erikn69:
Closes #1568

barryvdh:
Fixed with https://github.com/barryvdh/laravel-debugbar/pull/1570

PR #1576, opened 2024-03-13
Unserialize cache data

erikn69:
On redis driver, cache comes serialized
image
After `unserialize`
image


systemsolutionweb (approved):

barryvdh:
Hmmm I'm not a fan of unserialize. Are we 100% sure the value is serialized by the driver? Because otherwise a string posing as a serialized string could lead to issues (eg https://www.invicti.com/blog/web-security/untrusted-data-unserialize-php/ and more)

erikn69:
then it is better not to risk it

PR #1580, opened 2024-03-20
Remove deprecated TwigCollector

erikn69:
`TwigCollector` is deprecated 
[maximebf/php-debugbar/src/DebugBar/Bridge/Twig/TwigCollector.php#L30](https://github.com/maximebf/php-debugbar/blob/6cfc9ae6834adc9ac45f3bc81f1e6048a3d08103/src/DebugBar/Bridge/Twig/TwigCollector.php#L30)

### DIFF
--- a/src/DataCollector/ViewCollector.php
+++ b/src/DataCollector/ViewCollector.php
@@ -3,10 +3,12 @@
 namespace Barryvdh\Debugbar\DataCollector;
 
 use Barryvdh\Debugbar\DataFormatter\SimpleFormatter;
-use DebugBar\Bridge\Twig\TwigCollector;
+use DebugBar\DataCollector\AssetProvider;
+use DebugBar\DataCollector\DataCollector;
+use DebugBar\DataCollector\Renderable;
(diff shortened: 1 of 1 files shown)


barryvdh:
Hmm tests seem not to happy about this.

erikn69:
Works on a real app, but for tests, that path not exists

base_path has
- vendor
. laravel-debugbar
. php-debugbar

But on test
- laravel-debugbar
 . vendor/php-debugbar

Maybe using `ReflectionClass`, I'm going to try it

barryvdh:
Do we need the path? In the other collectors that isn't used?

erikn69:
>Do we need the path? In the other collectors that isn't used?

Yes, Extends TwigCollector only because original php-debugbar path cannot be accessed, 
It would be better if there was a way to declare that it should take the original path and not the overwritten one, like a flag

erikn69:
>Do we need the path? In the other collectors that isn't used?

You are right, not needed, At first it gave me a problem, but maybe it was the browser

PR #1585, opened 2024-03-22
Ability to add a custom stacktrace parsing method

OzanKurt:
Hello there,

I use https://github.com/yajra/laravel-datatables package in almost all of my projects.

I currently have a problem of seeing the source of the queries inside my DataTable classes.

For example:

I have a query builder setup inside my `UsersDataTable`.

```php
    public function query()
    {
        return User::query();
    }
```

Since this function only starts to bulid the query it doesn't show up in `debug_backtrace()`.

For this I've added a solution which allows me to check the "query sender" object and modify how the `$frame` was filled.

I've added simple static method to add my custom callables to `QueryCollector`, in there I can modify and return the `$frame` as I wish in case the `$trace['object']` is matching my first parameter `$objectType`.

Here is an example `DataTable::class` query tracer.

```php
QueryCollector::addCustomFrameParser(DataTable::class, function (object $frame, array $trace) {
    // Remove the 'App\' prefix from the namespace
    $relativeNamespace = str_replace('App\\', '', $trace['object']::class);

    // Convert namespace to a directory path
    $directoryPath = str_replace('\\', DIRECTORY_SEPARATOR, $relativeNamespace);
    $filePath = app_path($directoryPath) . '.php';

    // Get the file path inside the project from the class object
    $frame->file = $filePath;

    return $frame;
});
```

After I add this to my `AppServiceProvider`, I can simply see the change inside my "Query" tab.

Before:
image

After:
image

This way I can simply tell where I built the query and continue debugging/development.

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -34,6 +34,7 @@ class QueryCollector extends PDOCollector
         '/vendor/october/rain',
         '/vendor/barryvdh/laravel-debugbar',
     ];
+    protected static $customFrameParsers = [];
 
     /**
      * @param TimeDataCollector $timeCollector
@@ -41,6 +42,14 @@ class QueryCollector extends PDOCollector
     public function __construct(TimeDataCollector $timeCollector = null)
     {
         $this->timeCollector = $timeCollector;
+    }
+
+    /**
+     * Add a custom frame parser for a specific object type.
+     */
+    public static function addCustomFrameParser(string $objectType, callable $customFrameParser): void
+    {
+        static::$customFrameParsers[$objectType] = $customFrameParser;
     }
 
     /**
@@ -333,6 +342,15 @@ class QueryCollector extends PDOCollector
             !$this->fileIsInExcludedPath($trace['file'])
         ) {
             $frame->file = $trace['file'];
+
+            foreach (static::$customFrameParsers as $objectType => $customFrameParser) {
+                if (isset($trace['object']) && is_a($trace['object'], $objectType)) {
+                    $frame->line = '?';
+                    $frame = $customFrameParser($frame, $trace);
+                    $frame->name = $this->normalizeFilePath($frame->file);
+                    return $frame;
+                }
+            }
 
             if (isset($trace['object']) && is_a($trace['object'], 'Twig_Template')) {
                 list($frame->file, $frame->line) = $this->getTwigInfo($trace);


parallels999:
Did you try adding `PATH_TO_FILE/DataTablesController.php` to `options.db.backtrace_exclude_paths`?
I have a bridge query builder, and I had a similar problem,
I added the `file_path` to the `backtrace_exclude_paths` and I already got the correct file

https://github.com/barryvdh/laravel-debugbar/blob/03448e59739181179697dd996f581404db4ce7c5/config/debugbar.php#L213

OzanKurt:
Yes I tried but since the query is NOT actually sent from the `UsersDataTable` PHP's `debug_backtrace()` only detects the call from: `DataTablesController.php`

Also here is how mine looks like usually:

```php
'backtrace_exclude_paths' => [
    '/vendor/yajra/laravel-datatables',
    '/vendor/staudenmeir',
],
```

PaolaRuby:
Post `Before\After` with backtrace
image

OzanKurt:
Before:
image


After:
image

PaolaRuby:
extends `UsersDataTable` some yajra class?? or uses some yajra trait?

OzanKurt:
It extends the `DataTable` class from the package.

PaolaRuby:
I think it could be made more native, 
maybe some "demo minimal app" to find another approach?

OzanKurt:
I can make a demo next week, the change is actually very minimal anyway.

OzanKurt:
I actually already pushed a demo to GitHub.

https://github.com/OzanKurt/debugbar-demo

You can simply pull and run with `php artisan serve`.

erikn69:
This removes a step from the trace, I'm going to try another approach

OzanKurt:
So you suggest adding an extra source to the trace instead of modifying its file...

I'm fine with that approach as well.

One thing to add maybe would be the custom source parsing:

So instead of
```php
if (($trace['class'] ?? null) == 'Yajra\DataTables\Services\DataTable') {
```
looping through an array of customs would be nicer
```php
foreach (static::$customSourceParsers as $objectType => $customSourceParser) {
    if (isset($trace['object']) && is_a($trace['object'], $objectType)) {
```

erikn69:
>QueryCollector::addCustomFrameParser(DataTable::class, function (object $frame, array $trace) {
>    // Remove the 'App\' prefix from the namespace
>    $relativeNamespace = str_replace('App\\', '', $trace['object']::class);

I try to avoid that, not everyone has the knowledge to do that customization, and not all cases can be solved in this way
the way I add it would be ready to use
but @barryvdh decides

OzanKurt:
Why not add both, so people with the knowledge to customize can do it without needing a PR process.

barryvdh:
I'm not really sure about this, not seeing this in my own datatables. Don't want to complicate the backtrace too much.

barryvdh:
I'm not really against this, but should it match on the object type? Or just run and return a result if found?

PR #1594, opened 2024-03-31
Revert "Use dark theme by default"

barryvdh:
Reverts barryvdh/laravel-debugbar#1546

It seems not so popular, so I'll consider reverting it back the 'auto'.

What do you prefer? Thumbs up for auto, down for dark.


barryvdh:
Yeah but I think for debugbar it's usually different than the error page. I think it makes a bit sense to just follow the browser.

theKalash:
Thank you! Having to publish config files for every project was quite annoying.

Issue #632, opened 2017-05-11
Debugbar not displayed on some routes


damienfern:
Hi,

Debugbar isn't displayed on some views and i dont know why it doesn't :( . I'm sure i missed something but couldn't figure it out myself. Debugbar isn't displayed on 'rfq/newRevision' and on 'rfq/{n}/edit'. On others, it just works fine.

routes.php
```
<?php

/*
|--------------------------------------------------------------------------
| Web Routes
|--------------------------------------------------------------------------
|
| Here is where you can register web routes for your application. These
| routes are loaded by the RouteServiceProvider within a group which
| contains the "web" middleware group. Now create something great!
|
*/

Route::get('/', 'HomeController@index');
Route::get('lang/{lang}', ['as'=>'lang.switch', 'uses'=>'LanguageController@switchLang']);
Route::get('rfq/uploadOffer/{id_rfq}', 'RfqController@showUploadOfferForm')->name('rfq.uploadOffer');
Route::post('rfq/uploadOffer/{rfq}', 'RfqController@uploadOffer');
Route::get('rfq/newRevision/{id_rfq}', 'RfqController@showNewRevisionForm')->name('rfq.newRevision');
Route::post('rfq/newRevision/{rfq}', 'RfqController@newRevision');
Route::resource('rfq', 'RfqController');
Route::resource('offer', 'OfferController');
Route::get('offer/create/{id_rfq}', 'OfferController@create')->name('offer.create');
Route::resource('document', 'DocumentController');

Auth::routes();
```

app.php
```
<?php

return [

    /*
    |--------------------------------------------------------------------------
    | Application Name
    |--------------------------------------------------------------------------
    |
    | This value is the name of your application. This value is used when the
    | framework needs to place the application's name in a notification or
    | any other location as required by the application or its packages.
    */

    'name' => env('APP_NAME', 'GNMS Intranet'),

    /*
    |--------------------------------------------------------------------------
    | Application Environment
    |--------------------------------------------------------------------------
    |
    | This value determines the "environment" your application is currently
    | running in. This may determine how you prefer to configure various
    | services your application utilizes. Set this in your ".env" file.
    |
    */

    'env' => env('APP_ENV', 'production'),

    /*
    |--------------------------------------------------------------------------
    | Application Debug Mode
    |--------------------------------------------------------------------------
    |
    | When your application is in debug mode, detailed error messages with
    | stack traces will be shown on every error that occurs within your
[...]

PR #1599, opened 2024-04-01
Upstream twig extensions

erikn69:
It would be an idea to remove a lot of code, and php-debugbar already left behind the old syntax
#632 
https://github.com/rcrowe/TwigBridge/blob/4c256648cdc436fdd223e08ab1b8605b6fe75896/composer.json#L18-L19


barryvdh:
Is the compatibe with the old version? Or are user changes required.

erikn69:
It should be fully compatible, 
I gave it a quick check with https://github.com/rcrowe/TwigBridge and it didn't give me any problems

barryvdh:
Hmm I don't it always works, because of the order. The collectors are not added yet in the case I'm trying. See https://github.com/barryvdh/laravel-debugbar/pull/1600

erikn69:
>Hmm I don't it always works, because of the order. The collectors are not added yet in the case I'm trying.

I assume that the problem was already there before this change, or did this change cause it?

PR #1601, opened 2024-04-01
UI refinements

nckrtl:
This PR contains a lot of little refinements for the ui as discussed in https://github.com/barryvdh/laravel-debugbar/issues/1593. As I was going through the queries tab and applying changes for the spacings I noticed I couldn't let the other tabs untouched as the changes also affected those.

So I went through all the tabs and made sure everything is well aligned there as well. This results in the PR being a bit hard to review by looking at the CSS changes. Probably better to checkout the branch and compare side by side. I know you mentioned to also apply changes in the base package if applicable but I haven't checked that yet. If you like the changes in this PR I could take another look to see if anything from this PR could be extracted to the base package.

Below are screenshots of dark mode but I made sure light mode also renders correctly.

### 

Screenshot 2024-04-01 at 23 07 49

### 
The percentages bars were being displayed with table borders but I thought it would make sense to diplay them in light/dark background as the lines in Requests tabs are displayed as well. Making it a bit more consistent.

Screenshot 2024-04-01 at 23 08 08

### 
Updated spacing and line height of the lines in the error for bit better readability

Screenshot 2024-04-01 at 23 08 56

### 

Screenshot 2024-04-01 at 23 09 33

### 

Screenshot 2024-04-01 at 23 09 47

### 
I changed the color of the 1px line of the timeline measurement to blue to match the color that is being used on the timeline tab, felt more consistent like this. Also line height has been increased slightly for the table that contains the backtrace. Probably best to compare side by side

Screenshot 2024-04-01 at 23 10 12

### 
On the requests tab the background of the key items was not extending to the bottom of the of value items. This is now fixed and look a bit cleaner I think.

Screenshot 2024-04-01 at 23 10 47

### 
Adjust colours slightly and the table is moved up slightly to reduce the gap between the filters and the table

Screenshot 2024-04-01 at 23 11 03

### 
Slight spacing improvements and removed paddings. I could make it so that only the table will scroll and not the whole modal/popup content but will need to make a change in the base package I think.

Screenshot 2024-04-01 at 23 11 18

If there is anything you don't like let me know and I'll take a look.

PR #1603, opened 2024-04-03
Align SVG items

barryvdh:
Uses icons from https://github.com/maximebf/php-debugbar/pull/645 but with different fill in dark mode only.

image


src/Resources/laravel-debugbar.css
```diff
@@ -226,10 +226,6 @@ div.phpdebugbar-header {
...
-}
```
nckrtl, line 229: Could this change be unintended? I think it removes the Laravel logo:

Screenshot 2024-04-03 at 10 03 57

barryvdh: Whoops

barryvdh:
Not sure if this is need though, is it already included in https://github.com/barryvdh/laravel-debugbar/pull/1601 ?

PR #1604, opened 2024-04-03
Fix css

erikn69:
https://github.com/barryvdh/laravel-debugbar/pull/1601#discussion_r1549870141


nckrtl:
This middle aligns the icons better nicely for me yeah, but I think you still have those rather large top and bottom padding per line I think? That's probably caused by the line-height you mentioned in the comment.

I added the line height to make the request rows breath somewhat more, before:
Screenshot 2024-04-03 at 17 59 31

after:
Screenshot 2024-04-03 at 17 58 55

PR #1605, opened 2024-04-03
Cleanup + fix for regression since UI refinement branch

nckrtl:
- Removed invalid width calculation. I assumed it fixed the aligningment of the dt element, but as @parallels999 mentioned here https://github.com/barryvdh/laravel-debugbar/pull/1601#discussion_r1550017214 its invalid CSS and doesn't do anything. The alignment must've been fixed by something else.

- Brought back the padding of the search bar container in the messages tab to prevent overlap on the scrollbar. See https://github.com/barryvdh/laravel-debugbar/pull/1601#discussion_r1550033594Screenshot 2024-04-03 at 18 39 27

- Merged duplicate selector `dl.phpdebugbar-widgets-kvlist dt`

### DIFF
--- a/src/Resources/laravel-debugbar.css
+++ b/src/Resources/laravel-debugbar.css
@@ -313,10 +313,6 @@ dl.phpdebugbar-widgets-kvlist dd.phpdebugbar-widgets-value.phpdebugbar-widgets-p
     background: transparent;
 }
 
-dl.phpdebugbar-widgets-kvlist dt {
-    width: calc(25%-10px);
-}
-
 dl.phpdebugbar-widgets-kvlist dd {
     margin-left: 25%;
 }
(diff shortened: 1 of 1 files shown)


erikn69:
Almost, but i can handle it now 👍
image

nckrtl:
Hmm, that would trigger my OCD still :) more than happy to reduce the width more. Seems it would need another trim of 10px on either side.

src/Resources/laravel-debugbar.css
```diff
@@ -313,10 +313,6 @@ dl.phpdebugbar-widgets-kvlist dd.phpdebugbar-widgets-value.phpdebugbar-widgets-p
     background: transparent;
 }
 
-dl.phpdebugbar-widgets-kvlist dt {
-    width: calc(25%-10px);
```
erikn69, line 317: after this was removed
image
now you can no longer read the model names

PR #1610, opened 2024-04-04
Toolbar refinements

nckrtl:
Was taking another look at the toolbar and found some more things that could be improved
- Different paddings were being used for left and right side, this is now consistent
- Weird hover state on active tab/element
- Text appears bit fuzzy because of text shadow, I think its more clear to show it without a text shadow
- Icons on the right side seem to have different paddings, made those consistent as well
- applied better vertical alignment for the icons used on the right side

Visual side by side:
visual-changes

Also noticed that the rows of the request data is not nicely aligned anymore:
Screenshot 2024-04-04 at 21 39 23
I saw this remark: https://github.com/barryvdh/laravel-debugbar/pull/1605#discussion_r1551759026 and also the PR that is causing the misalignment: https://github.com/barryvdh/laravel-debugbar/pull/1608

This CSS fixes the alignment, while making sure there is enough space for the key to be fully displayed and not being cut off. 
```
dl.phpdebugbar-widgets-kvlist dt {
    width: 25%;
    width: calc(25% - 10px);
}
```
Note that this is now properly calculating the width instead of `width: calc(25%-10px);` which I included and then removed in some commits ago as it was invalid due to missing spaces.

Screenshot 2024-04-05 at 00 06 13

### DIFF
--- a/src/Resources/laravel-debugbar.css
+++ b/src/Resources/laravel-debugbar.css
@@ -316,6 +316,10 @@ dl.phpdebugbar-widgets-kvlist dd.phpdebugbar-widgets-value.phpdebugbar-widgets-p
     background: transparent;
 }
 
+dl.phpdebugbar-widgets-kvlist dt {
+    width: calc(25% - 10px);
+}
+
 dl.phpdebugbar-widgets-kvlist dd {
     margin-left: 25%;
 }
--- a/src/Resources/laravel-debugbar.css
+++ b/src/Resources/laravel-debugbar.css
@@ -795,7 +795,11 @@ div.phpdebugbar-header-right > * {
 }
 
 div.phpdebugbar-header-right > *:first-child {
-    border-right: none;
+    border-right: 0;
+}
+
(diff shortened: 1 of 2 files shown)


src/Resources/laravel-debugbar.css
```diff
@@ -266,19 +271,20 @@ a.phpdebugbar-restore-btn {
 }
 
 div.phpdebugbar-header > div > * {
+    padding: 5px 8px;
     font-size: 13px;
-    padding: 5px;
+    height: 20px;
```
barryvdh, line 276: This fixed height gave me problems before, is this required?

nckrtl: Changes applied in the base package through this merged PR https://github.com/maximebf/php-debugbar/pull/647/files should make sure it works well.

Screenshot 2024-04-05 at 14 05 05

The height of 16px is now being set to 20px as the minimum height of `div.phpdebugbar-header` in this package is  30px, compared to a min-height of 26px in the base package.

side-by-side

PR #1618, opened 2024-04-20
Adding response data to RequestCollector Widget/Tab

HassanDomeDenea:
In the debugbar, I wanted to see the response data, so that I don't need the devtools network tab.


barryvdh:
Yeah think this will increase the size a lot.

PR #1624, opened 2024-06-04
Fix route path in phar

murrant:
Don't call realpath() on route file path, it breaks routes inside phars.
It doesn't match the $configPath above it.  I can't figure out why realpath() would be called here.

Don't ask why I'm using in a phar ;)


parallels999:
> Don't ask why I'm using in a phar ;)

you read my mind 😄😄😄

PR #1630, opened 2024-07-08
Dragging Feature

devzkhalil:
> I'm a die-hard fan of the Laravel Debugbar package!

**The reason for this PR!**

image

In the above image, you can see that I have a menu in the same place as the Debugbar area. Whenever I try to click on the menu, I face issues with the click! I know many people face this problem and may face it in the future!

Therefore, I have tried to implement a feature here that allows the Debugbar to be dragged!

https://www.loom.com/share/438c76dec6a74fbd9f62063ce75f9a5c

### DIFF
--- a/src/Resources/cache/widget.js
+++ b/src/Resources/cache/widget.js
@@ -56,4 +56,88 @@
         }
     });
 
+    // ============= Handling Dragging Behavior ============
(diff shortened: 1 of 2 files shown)


parallels999:
Why should this functionality be located in the cache widget?

maybe you should do an upstream to main package: [maximebf/php-debugbar](https://github.com/maximebf/php-debugbar)

PR #1631, opened 2024-07-09
Add (initial) support for PHP 8.4

jnoordsij:
This MR:
- Runs the testsuite against PHP 8.4 (uses the currently still unstable beta version); to allow this the `--ignore-platform-req=php+` is added to `composer` calls in integration tests, to ignore checking package bounds on upper PHP versions
- ~~Replaces deprecated implicit nullable types in parameter declarations (see https://wiki.php.net/rfc/deprecate-implicitly-nullable-types)~~ see #1632

To make (integration) tests pass, the following changes are introduced:
- As Laravel cannot install yet by default with PHP 8.4, per https://github.com/nette/schema/issues/68, the suggestion there is applied to ignore PHP version in platform req checks for composer. Although opting to wait for a fix of the mentioned issue is also viable, I think not checking PHP version (through composer) for integration tests is a reasonable change, as any impactful change should already be breaking the tests anyways, and still end-users will be warned when installing with composer.
- The integration test is somewhat altered, as previously it checked for an empty `logs` directory. Due to deprecations in previous Laravel/Lumen versions being triggered on PHP 8.4, a log file is created by Laravel to log these messages. This MR changes the integration test to solely test functionality of the clear command, by adding a test file to the directory that should be cleared and subequently testing emptiness of that directory.

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -18,7 +18,7 @@ jobs:
             COMPOSER_NO_INTERACTION: 1
         strategy:
             matrix:
-                php: [8.3, 8.2, 8.1, 8.0]
+                php: [8.4, 8.3, 8.2, 8.1, 8.0]
                 lumen: [10.*, 9.*]
                 exclude:
                   - lumen: 10.*
@@ -66,7 +66,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [8.3, 8.2, 8.1, 8.0]
+                php: [8.4, 8.3, 8.2, 8.1, 8.0]
                 laravel: [11.*, 10.*, 9.*]
                 exclude:
                   - laravel: 10.*
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -54,10 +54,11 @@ jobs:
             - name: Execute clear run
               run: |
                   cd sample
+                  mkdir -p "storage/debugbar/" && touch "storage/debugbar/foo.json"
                   php artisan debugbar:clear
             - name: Check file count in logs
               run: |
-                if [ `ls -1q "sample/storage/logs/" | wc -l` -gt 0 ];then exit 1;fi
+                if [ `ls -1q "sample/storage/debugbar/" | wc -l` -gt 0 ];then exit 1;fi
     php-laravel-integration-tests:
         runs-on: ubuntu-latest
         timeout-minutes: 15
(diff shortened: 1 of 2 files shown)


jnoordsij:
Marked as draft as PHP 8.4 for tests might be just a little too early. The deprecation fixes were moved to #1632 so they can be merged already.

jnoordsij:
@barryvdh I've slightly reworked integration tests to make sure everything passes, LMK what you think!

barryvdh:
Why do we need to ignore the platform reqs?

jnoordsij:
As long as `nette/schema` does not tag 8.4 compatible (see https://github.com/nette/schema/issues/68), composer won't install Laravel on 8.4+.

barryvdh:
I think it's tagged now, can we remove the platform reqs skipping?

.github/workflows/run-integration-tests.yml
```diff
@@ -39,25 +39,26 @@ jobs:
...
               run: sed -e 's|// \$app->register(App\\\Providers\\\EventServiceProvider::class);|&\n$app->register(\\Barryvdh\\Debugbar\\LumenServiceProvider::class);|' -i sample/bootstrap/app.php
 
             - name: Execute clear run
               run: |
                   cd sample
+                  mkdir -p "storage/debugbar/" && touch "storage/debugbar/foo.json"
                   php artisan debugbar:clear
             - name: Check file count in logs
               run: |
-                if [ `ls -1q "sample/storage/logs/" | wc -l` -gt 0 ];then exit 1;fi
```
barryvdh, line 60: What is this change exactly?

jnoordsij:
> I think it's tagged now, can we remove the platform reqs skipping?

Just rebased and removed the commit, although I'd still argue it might've been beneficial to have for testing either way.

.github/workflows/run-integration-tests.yml
jnoordsij: I'm not exactly sure on what the intended check was doing. This check was introduced with #1118 and seems to aim to check for 'succesful completion' of the `php artisan debugbar:clear` command on a fresh Laravel installation. Here 'succesful' seems to be measured by "not logging any messages".

However, this metric seems slightly off to me. In this case it was erring, because (due to https://github.com/maximebf/php-debugbar/pull/687) deprecations are being logged while running the command. There are ways to work around this, e.g. suppressing deprecations with some Laravel flag.

I've here opted to rather than trying to patch the old definition of 'succes', come up with a new one that actually feature-tests the actual command. Given that the purpose of the command is cleaning json files from the `storage/debugbar` directory, I've now added a check to ensure the command does just this. Thus the new definition of 'succesful' is now the combination of `php artisan debugbar:clear` exits with code 0 and does feature-wise what it's expected to do.

PR #1656, opened 2024-09-09
Use laravel logo icon on LaravelCollector

erikn69:
image

image

maybe bigger with `phpdebugbar-fa-lg`?

image

### DIFF
--- a/src/DataCollector/LaravelCollector.php
+++ b/src/DataCollector/LaravelCollector.php
@@ -49,7 +49,7 @@ class LaravelCollector extends DataCollector implements Renderable
     {
         return [
             "version" => [
-                "icon" => "github",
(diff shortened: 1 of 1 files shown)


src/DataCollector/LaravelCollector.php
```diff
@@ -49,7 +49,7 @@ public function getWidgets()
     {
         return [
             "version" => [
-                "icon" => "github",
+                "icon" => "laravel phpdebugbar-fab",
```
barryvdh, line 52: why the phpdebugbar-fab?

erikn69: After FA 4.x, icons were divided into several sets, example, `fab`(_`phpdebugbar-fab` here_), what does font awesome brands mean

image

PR #1657, opened 2024-09-10
Fix redirect session

barryvdh:
Get the response early for redirects, and get the stacked data also early.

Since some time the redirects are not stacked anymore, I think because the session is saved before the debugbar is injected.
This moves the session stuff earlier in the cycle. It will continue for the regular request.


barryvdh:
Ugh, not really sure why tests are failing on pipelines, but working locally.

Anyway, would be nice if someone could test this on their projects.

PR #1659, opened 2024-09-10
Avoid js `alert` on copy, and avoid params opening on link click

erikn69:
Same functionality as

- https://github.com/maximebf/php-debugbar/pull/676 (_remove js `alert`_)
- https://github.com/maximebf/php-debugbar/pull/667


barryvdh:
Oh whoops, meant to do that in https://github.com/barryvdh/laravel-debugbar/pull/1658 but missed the alert. Thanks.

PR #1663, opened 2024-09-11
Avoid `Undefined variable $sessionManager`

erikn69:
image


erikn69:
@barryvdh hi, #1662 seems to break ajax capture too

barryvdh:
Yeah probably going to revert that.

barryvdh:
See https://github.com/barryvdh/laravel-debugbar/pull/1666
Which case is for ajax?

erikn69:
ajax  seems to be working👍

PR #1672, opened 2024-09-17
Correctly align and space comments

u01jmg3:
[As per other Laravel configs](https://github.com/laravel/laravel/blob/11.x/config/auth.php#L5-L14), correctly align and space those used in the Debugbar config


barryvdh:
Closing for now, want to change the config.

PR #1675, opened 2024-09-20
Sql copy button at line start

erikn69:
I don't see much usability to put the button at the end of the query,
when the queries are very long it is annoying to go down and look for the button,
besides the position depends on the length of the last line of the query, the button place is being variable.

image

### DIFF
--- a/src/Resources/laravel-debugbar.css
+++ b/src/Resources/laravel-debugbar.css
@@ -927,6 +927,7 @@ pre.phpdebugbar-widgets-code-block ul.phpdebugbar-widgets-numbered-code li {
 
 div.phpdebugbar-widgets-sqlqueries span.phpdebugbar-widgets-copy-clipboard {
     float: none !important;
+    margin-left: 0 !important;
 }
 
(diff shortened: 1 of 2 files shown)


tpetry:
The idea was that the copy button is not the most important thing. And when the copy button is left to the text than you have "jumping" text: "transaction start" is completely left, and queries are slightly intended. thats a small ux issue.

erikn69:
>copy button is not the most important thing

It's the option that I use the most 😄
So is there possibility of returning it to its original position?
I have an app with quite large SQL, and I copy them to a formatter that creates a graph,
it helps me optimize or fix the queries

barryvdh:
Hmm I think it probably looks better on the left indeed. Or completely to the right where the other buttons are?

barryvdh:
On php-debugbar it's part of the rest of the buttons: 
image

erikn69:
>  completely to the right where the other buttons are?

It was originally like that, but i was changed
It can be on the left or right, the important thing is that it is easy to find

barryvdh:
With larger queries the buttons jump around a bit anyways, so not really sure.

PR #1677, opened 2024-09-23
Fix ErrorException on empty responses

stefket:
For example, the spatie/laravel-honeypot package returns an empty response if the honeypot field is filled in. In this case, the Laravel Debugbar triggers an ErrorException with the message “Uninitialized string offset 0.”

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -905,7 +905,7 @@ class LaravelDebugbar extends DebugBar
 
         // Check if content looks like JSON without actually validating
         $content = $response->getContent();
-        if ($content !== false && in_array($content[0], ['{', '['], true)) {
+        if ($content && in_array($content[0], ['{', '['], true)) {
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


src/LaravelDebugbar.php
```diff
@@ -905,7 +905,7 @@ protected function isJsonRequest(Request $request, Response $response)
 
         // Check if content looks like JSON without actually validating
         $content = $response->getContent();
-        if ($content !== false && in_array($content[0], ['{', '['], true)) {
+        if ($content && in_array($content[0], ['{', '['], true)) {
```
erikn69, line 908:
```diff
-if ($content && in_array($content[0], ['{', '['], true)) {
+if (is_string($content) && in_array($content[0], ['{', '['], true)) {
```

SocolaDaiCa:
```diff
-if ($content && in_array($content[0], ['{', '['], true)) {
+if (is_string($content) && json_decode($content) !== null) {
```

I think we should use json_decode because may be json has contains space first or last

barryvdh: https://github.com/barryvdh/laravel-debugbar/pull/1677

PR #1678, opened 2024-09-23
chore: better results for postgres visual explain

ImJustToNy:
As per:
{80C8ADC5-DCA5-411C-962F-0B7F0F1773C0}
At https://explain.dalibo.com/

### DIFF
--- a/src/Support/Explain.php
+++ b/src/Support/Explain.php
@@ -105,7 +105,7 @@ class Explain
     {
         return (string) Http::asForm()->post('https://explain.dalibo.com/new', [
             'query' => $query,
-            'plan' => $connection->selectOne("EXPLAIN (FORMAT JSON) {$query}", $bindings)->{'QUERY PLAN'},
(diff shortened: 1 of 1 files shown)


barryvdh:
@tpetry thoughts?

tpetry:
Using the ANALYZE mode is ok too. I didnt initially use it as it executes the query again (and slower). With the old approach it would have made the debugbar really slow.

with the new approach its feasible. But when a query takes 30s to execute, the user has to wait those 30s. And there is no indication in the UI that something is still working. From a UX perspective it would be weird and in my opinion the UI needs to changed to reflect that something is working for so long.

tpetry:
Another issue is that `ANALYZE` really executes the query. The current implementation only checks whether the `select` keyword is included in the query to allow it for analyzing. But data manipulation queries involving a subquery (`UPDATE example set col = (SELECT ...)`) would also be included in that simple check.

This is not a problem for the EXPLAIN statement that doesn't run the statement. But with `ANALYZE` the behaviour may be weird (or produce data inconsistency issues) when the same data manipulation is executed multiple times.

PR #1687, opened 2024-10-09
Add Dusk tests for Queries tab

mccahan:
**NOTE: This requires the regression test fix from PR #1686 before it will pass**

Adds testing that would cover the thing I broke:
- Confirms the Queries tab is counting
- Confirms that a test query appears
- Confirms clicking the query expands it and shows bindings and trace

The two screenshots it grabs:

queries-expanded
queries-tab


barryvdh:
Can you merge master? It seems I don't have permissions.

mccahan:
Are you able to re-run the failed checks? They should fire now with that other PR merged. If you can't, I should just be able to force push this one

barryvdh:
Yeah I tried to run them but didn’t work. Thanks!

Issue #1689, opened 2024-10-17
Error message flooding my laravel.log


avn-dev:
In combination with laravel-pulse this is getting spammed in my logfile (when I'm viewing a /pulse resource):
=> most recent laravel, pulse & debugbar version installed
database: sqlite

[2024-10-17 03:32:52] Debugbar exception: Strings with null bytes cannot be escaped. Use the binary escape option.

barryvdh:
Seems similar to https://github.com/laravel/telescope/issues/1489 so probably related to use also doing substituteBindingsIntoRawSql

I tried https://github.com/barryvdh/laravel-debugbar/pull/1690 as fix

PR #1690, opened 2024-10-17
Add try/catch for the substituteBindings

barryvdh:
Fixes https://github.com/barryvdh/laravel-debugbar/issues/1689

Not sure if we can detect/prevent this though. 

Reproducable case:
- Install new laravel/laravel with sqlite
- Install Pulse https://laravel.com/docs/11.x/pulse#installation
- Go to /pulse


barryvdh:
@tpetry any ideas about this? I see you replied to https://github.com/laravel/telescope/issues/1489 also

tpetry:
The change looks good and I would have suggested the same. I've heard that the escaping logic sometimes fails but no one ever came up with an example. Its always reports like the Telescope one that there is one problem, but instead of sharing a reproduceable case they fix their issue and call it finished 🤷‍♂️

I didn't know that with pulse there is a reproduceable case. I'll look into that next week - currently ill.

PR #1573, opened 2024-03-12
Make sure session is saved after changing

barryvdh:
Looks like the session isn't properly saved. Not sure if this is Laravel 11 specific or not.
Happens when redirecting.

PR #1710, opened 2024-12-23
Add middleware to web to save session

### LINKED ISSUE #1704: A lot of issues with not working addMessage in stale state and never solved
I'm having this issue that are also related to this:

- https://github.com/barryvdh/laravel-debugbar/issues/1154
- https://github.com/barryvdh/laravel-debugbar/issues/1025
- https://github.com/barryvdh/laravel-debugbar/issues/905
- https://github.com/barryvdh/laravel-debugbar/issues/831
- https://github.com/barryvdh/laravel-debugbar/issues/177

I'm having the same issue trying to force to appear a message to give visibility to an improperly controlled error.

In my case, I found that a FormRequest with an improperly validation, can cause not pass a validation and not see the error. I think to show it on debugbar for get the needed feedback that what is happening.

I try to do a Middleware to catch the error, dd show the message that debugbar:addMessage doesn't show.
I try to do another class to extends, that have failedValidation, and the same problem. And in this point I try some other things.

This is my final test, that seems to show a non expected behaviour:

```
<?php

namespace App\Http\Requests;

use Barryvdh\Debugbar\Facades\Debugbar;
use Exception;
use Illuminate\Contracts\Validation\Validator;
use Illuminate\Foundation\Http\FormRequest;


class FormRequestValidation extends FormRequest
{
    /**
     * Manejar la validación fallida.
     */
    protected function failedValidation(Validator $validator)
    {
        // Registrar errores en Debugbar
        Debugbar::addMessage($validator->errors()->toArray(), 'Validation failed');

        // throw new Exception('Validation failed: ' . json_encode($validator->errors()->toArray(), JSON_PRETTY_PRINT));

        // Llamar al comportamiento predeterminado (lanza ValidationException)
        parent::failedValidation($validator);
    }
}
```

With the throw commented, I can't see messages on debugbar.
With the throw uncommented, I can see TWO messages on debugbar.

To try to reproduce this problem, is only needed a form validation with a required field that is not submited.

What is the solution? I don't know, but a lot of people seems to have this kind of problem, but no solution with a problem that seems to happen from nearly 2014.

barryvdh:
The Web middlewaregoup contains the SaveSession middleware, which closes the session. This makes it impossible (as far as I can tell) to save to the session, eg. for redirects.

This will add the middleware to both the regular and the normal group. It will still only boot once and inject once because of the checks, but this should make redirects work again, without any of the workarounds. (#1666 #1662 #1657 #1613 #1591 #1574 #1573) and fixes #1704

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -136,5 +136,6 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
     {
         $kernel = $this->app[Kernel::class];
         $kernel->pushMiddleware($middleware);
+        $kernel->appendMiddlewareToGroup('web', $middleware);
     }
 }


shawe:
It doesn't seem enough for the case where I detected it.

My test case is this:

`
<?php

namespace App\Http\Requests;

use Barryvdh\Debugbar\Facades\Debugbar;
use Illuminate\Contracts\Validation\Validator;
use Illuminate\Foundation\Http\FormRequest;
use Illuminate\Support\Facades\Log;

class FormRequestValidation extends FormRequest
{
    protected function failedValidation(Validator $validator)
    {
        // Add a message to the Laravel Debugbar
        Debugbar::addMessage($validator->errors()->toArray(), 'Validation failed');
        // Add a message to the Laravel Log
        Log::debug('Validation failed: ' . json_encode($validator->errors()->toArray(), JSON_PRETTY_PRINT));

        // Call the default behavior (throws ValidationException)
        parent::failedValidation($validator);
    }
}
`

This must show the list of errors when a form is not validated here, but it's not showed.

image

barryvdh:
Does the dropdown menu not show up on the right side?

shawe:
> Does the dropdown menu not show up on the right side?

Yes, and it is in the last request, which should have and display an addMessage capturing the form validation errors.

This may be a very particular problem using Livewire, as it intercepts a thrown exception. I can't remember any other circumstances where I encountered the same problem without using Livewire, so I'll try to see if this fixes it in that case when I remember.

barryvdh:
In a redirect, it shows the latest request as last, so you should see the log messages on the first. is that the cases with this PR, or do you not see them at all?

shawe:
> In a redirect, it shows the latest request as last, so you should see the log messages on the first. is that the cases with this PR, or do you not see them at all?

In this case there is no redirection, the debugbar was showing the last request without problems, the problem it's apparently only with added messages from any type. I have tabs with no data disabled, so there are fewer visible options because they are empty.

barryvdh:
Is this livewire? Or why is there no redirect?

shawe:
> Is this livewire? Or why is there no redirect?

Yes, I say you here:

> > Does the dropdown menu not show up on the right side?
> 
> Yes, and it is in the last request, which should have and display an addMessage capturing the form validation errors.
> 
> This may be a very particular problem using Livewire, as it intercepts a thrown exception. I can't remember any other circumstances where I encountered the same problem without using Livewire, so I'll try to see if this fixes it in that case when I remember.

barryvdh:
Are you on the dev version?

barryvdh:
Can you try to update? I've made a fix

eusonlito:
@barryvdh fixed. Thanks!

PR #1717, opened 2025-01-12
[WIP] Use php-debugbar dark theme, move to variables

barryvdh:
This is pending on https://github.com/php-debugbar/php-debugbar/pull/703

- Move hardcoded colors to CSS variables
- Add a dark theme for php-debugbar, based on CSS variables
- Move hardcoded colors for laravel-debugbar to CSS variable overrides
- Tweak Laravel-debugbar darkmode to only use CSS variables.

This should make it easier to maintain both themes, if the existing variables are used. And a lot less CSS duplication.

Some color overrides should probably still be done in php-debugbar instead of here.

Note: This requires the feat-darktheme branch for debugbar to work.


barryvdh:
I think this is done, but requires the latest php-debugbar beta.

erikn69:
Hello, the opener modal has no theme colors, is this intentional?
image

barryvdh:
Hmm no, it should have. Did you do a hard refresh?

erikn69:
Of course i do that 😂
I notice this
image
I did add `data-theme="dark"` to the modal, and now it works
image
Maybe you forgot something on the first page load?

barryvdh:
Hmm might be a timing issue then. What is your initial version, light? Or auto with light? It should set either dark or light anyways, so I’ll have to check.

erikn69:
>might be a timing issue then

I think the same, it would be better to add it to the openhandler render
No light, always dark like vampires 👍,

erikn69:
would  It be possible to add something like `text-overflow: ellipsis;` in the openhandler link td?
it puts the entire URL and in the GET(big url) the horizontal scrollbar appears, it is annoying
image

maybe the header could be fixed

barryvdh:
Fixed the theme and some default styling here: https://github.com/php-debugbar/php-debugbar/pull/715 and removed some overrides in https://github.com/barryvdh/laravel-debugbar/pull/1723

PR #1726, opened 2025-02-01
Add more data to timeline

barryvdh:


src/ServiceProvider.php
```diff
@@ -48,42 +48,6 @@ function ($app) {
...
-
-                $extensions = array_reverse($factory->getExtensions());
-                $engines = array_flip($extensions);
-                $enginesResolver = $application->make('view.engine.resolver');
-
-                foreach ($engines as $engine => $extension) {
-                    $resolved = $enginesResolver->resolve($engine);
-
-                    $factory->addExtension($extension, $engine, function () use ($resolved, $laravelDebugbar) {
-                        return new DebugbarViewEngine($resolved, $laravelDebugbar);
```
parallels999, line 74: Now the view creation time is not measured.
[DebugbarViewEngine](https://github.com/barryvdh/laravel-debugbar/blob/master/src/DebugbarViewEngine.php) class now is no longer used, it should be removed

barryvdh: Yes this seemed a bit hacky. Should probably removed it indeed.

PR #1728, opened 2025-02-04
Preview action_name on request tooltip

angeljqv:


barryvdh:
So this is okay but as @erikn69 noticed, the tooltip is currently not refreshed. So we have to map them somehow.

barryvdh:
I think that should be solved with https://github.com/php-debugbar/php-debugbar/pull/724
I'll update it after this PR.

barryvdh:
Fixed with https://github.com/barryvdh/laravel-debugbar/pull/1732

PR #1732, opened 2025-02-13
Map tooltips

barryvdh:


src/DataCollector/RequestCollector.php
```diff
@@ -191,7 +186,19 @@ public function collect()
             $htmlData['telescope'] = '<a href="' . $url . '" target="_blank">View in Telescope</a>';
         }
 
-        return $htmlData + $data;
+        $tooltip = Arr::only($data, ['status', 'full_url']);
```
erikn69, line 189: This give me a problem, `full_url` could be an array, and it is impossible to read any data
image

https://github.com/barryvdh/laravel-debugbar/blob/67e1bbdac009a9b5c947bd9ed58a477fc36b2af4/src/DataCollector/RequestCollector.php#L138

PR #1736, opened 2025-02-21
Show request status as badge

barryvdh:
Screenshot 2025-02-21 at 10 51 54


parallels999:
For reasons of space, wouldn't it be better to just leave the code number? to something shorter

barryvdh:
Perhaps but it’s probably important

PR #1722, opened 2025-01-30
fix: The search box hides the horizontal scroll bar in the Logs.

uniho:
See the image below.

image

PR #1754, opened 2025-02-28
an alternative to PR #1722.

uniho:
This is an alternative to PR #1722.


uniho:
Please refer to the following for `{position: sticky; top: 100vh;}`.
https://css-tricks.com/a-clever-sticky-footer-technique/

barryvdh:
If this is still an issue, it should be fixed in php-debugbar.

PR #1762, opened 2025-03-20
Fix `capture_dumps` option on laravel `dd();`

parallels999:
Using laravel `dd();` it shows nothing, this fix it, 
Apparently there is no way to differentiate dumps from dds, since Laravel's dd first executes a `dump();` and then an `exit(1);`


barryvdh:
The default should be false. Can we change that back? See #1782

parallels999:
Yes, my bad, I didn't realize that.

PR #1764, opened 2025-03-25
Show estimate of cache byte usage

erikn69:
image

### DIFF
--- a/src/DataCollector/CacheCollector.php
+++ b/src/DataCollector/CacheCollector.php
@@ -28,6 +28,7 @@ class CacheCollector extends TimeDataCollector
         parent::__construct();
 
         $this->collectValues = $collectValues;
+        $this->memoryMeasure = true;
     }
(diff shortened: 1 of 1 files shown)


parallels999 (approved):

systemsolutionweb (approved):

ssw1cblarrion (approved):

miken32:
This seems to be breaking things for us. Trying to store a model in cache works fine in production, but this change is breaking my dev environment with a "Serialization of 'Closure' is not allowed" exception. This code should have been in a try/catch rather than assuming it will work. #1993 fixes this problem for us.

barryvdh:
I'm not sure if thats a valid case, because I got from the cache when trying to serialize a closure, not just from debugbar. But tweaked the check and added some tests in https://github.com/fruitcake/laravel-debugbar/pull/1994

Issue #1775, opened 2025-04-08
Recurrent warning in Snipe-IT logs due to null values in QueryCollector.php


Mathias-DS:
This is a fix for https://github.com/snipe/snipe-it/issues/16476 where a null value makes laravel-debug bar fill the log with warnings.

While running Snipe-IT on PHP 8.2.12, I encountered the following Laravel Debugbar warning:

LOG.warning: strpos(): Passing null to parameter https://github.com/snipe/snipe-it/issues/1 ($haystack) of type string is deprecated in snipe-it\vendor\barryvdh\laravel-debugbar\src\DataCollector\QueryCollector.php on line 381

This issue occurs because PHP 8.1 and later deprecated passing null values to strpos(). If either $class or $filename is null, strpos() throws a warning.

To fix this temporarily, I modified the code as follows:
Original Code (QueryCollector.php Line 381):

if (strpos($class, $filename) !== false) {

⬇
Fixed Version (Ensuring Null Safety):

if (!is_null($class) && !is_null($filename) && strpos($class, $filename) !== false) {

After applying this fix, Debugbar continued working without errors.

Suggested Fix:
In future updates, adding NULL checks before using strpos() will help ensure compatibility with PHP 8.1+.

System Details:
📌 Snipe-IT Version: v8.0.4 build 17196 (gc29bdbdac)
📌 PHP Version: 8.2.12
📌 Environment: Windows 11 / XAMPP

PR #1776, opened 2025-04-08
Update QueryCollector.php fix issue #1775

Mathias-DS:
Test for null value in strpos in QueryCollector.php to solve laravel-debugbar issue #1775 and snipe-it issue 16476.

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -378,7 +378,7 @@ class QueryCollector extends PDOCollector
         $filename = pathinfo($file, PATHINFO_FILENAME);
 
         foreach ($this->middleware as $alias => $class) {
-            if (strpos($class, $filename) !== false) {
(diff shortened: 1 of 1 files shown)


src/DataCollector/QueryCollector.php
```diff
@@ -378,7 +378,7 @@ protected function findMiddlewareFromFile($file)
         $filename = pathinfo($file, PATHINFO_FILENAME);
 
         foreach ($this->middleware as $alias => $class) {
-            if (strpos($class, $filename) !== false) {
+            if (!is_null($class) && !is_null($filename) && strpos($class, $filename) !== false) {
```
parallels999, line 381: Couldn't it be something simpler? like
```php
if ($class && strpos($class, (string) $filename) !== false) {
``` 
```suggestion
            if ($class && strpos($class, (string) $filename) !== false) {
```

PR #1779, opened 2025-04-09
Do not rely on DB::connection() to get information in query collector

### LINKED ISSUE #1778: Queries need a real database connection since v3.14.0
My application collects data not only from MariaDB, but also from ElasticSearch, Drupal (via GraphQL), Microsoft Dynamics CRM and Microsoft PowerAutomate.
Debugbar's query logs are incredible helpful, so I integrated all the adapters into Debugbar's "queries" tab by logging queries by dispatching ` \Illuminate\Database\Events\QueryExecuted` objects. I only needed a small dummy connection class that returned the database and driver names for that to work.

Now I'm upgrading from Laravel 8 to 9, and with it `laravel-debugbar` from v3.7.0 to v3.14.10, and I am seeing javascript errors in the debugbar code:
> data.statements is undefined

The reason is that an exception happens during debugbar processing:
> Debugbar exception: Database connection [graphql] not configured.

While previously the connection's name was simply taken from `$query['connection']` in `QueryCollector::collect()`, it asks Laravel's DatabaseManager for the connection and gets its name from there since commit ed37d1a4640ab3228795a380cf069bb277c5bd9d.

This requires me to configure and register the connection within laravel's database management code, which is pretty complex.

I'd like to see that the code is changed, so that the connection name is taken from the connection object that's provided by the query.

cweiske:
Instead directly use the query's own connection object.

This makes it possible to log queries for custom non-sql connections into debugbar relatively easily.

Related: ed37d1a4640ab3228795a380cf069bb277c5bd9d
This commit introduced the usage of DB::connection().

Resolves: https://github.com/barryvdh/laravel-debugbar/issues/1778

### DIFF
--- a/src/DataCollector/QueryCollector.php
+++ b/src/DataCollector/QueryCollector.php
@@ -196,7 +196,7 @@ class QueryCollector extends PDOCollector
             'time' => $time,
             'memory' => $this->lastMemoryUsage ? memory_get_usage(false) - $this->lastMemoryUsage : 0,
             'source' => $source,
-            'connection' => $query->connection->getName(),
+            'connection' => $query->connection,
             'driver' => $query->connection->getConfig('driver'),
             'hints' => ($this->showHints && !$limited) ? $hints : null,
             'show_copy' => $this->showCopyButton,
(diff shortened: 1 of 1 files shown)


cweiske:
Thank you for merging!

PR #1784, opened 2025-04-25
Make all scalar config values configurable through environment variables

wimski:
This PR makes all scalar values in `config/debugbar.php` configurable through environment variables. The original values are kept as defaults to be backwards-compatible. The naming of the environment variables always follows these rules:
* start with the package name
* use exact keys, including nesting
* use `_` as separator between parts
* uppercase all

This feature makes it easy to configure the debug bar for specific debugging situations, without having to make actual changes to the config file.

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -15,7 +15,7 @@ return [
      */
 
     'enabled' => env('DEBUGBAR_ENABLED', null),
-    'hide_empty_tabs' => true, // Hide tabs until they have content
+    'hide_empty_tabs' => env('DEBUGBAR_HIDE_EMPTY_TABS', true), // Hide tabs until they have content
     'except' => [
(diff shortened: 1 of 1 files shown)


barryvdh:
I guess okay but feels a bit 'overkill' maybe?

wimski:
Not everybody is going to use this, but it shouldn't get in the way if you don't.

I usually configure the debug bar for specific cases by disabling unwanted features (removing clutter) and tweaking settings for the specific area I'm debugging. And developers within a team might have a preference for different default settings. This update makes it easy to configure per such 'environment'.

You could also publish the config, change the values there and add the file to the `.gitignore`, but that removes the option to have shared defaults on a project level. And it feels a bit weird/dirty to ignore a (explicitly) published config file.

PR #1798, opened 2025-05-23
Better contrast in dark theme titles.

angeljqv:
In the light theme it is differentiated by the `bold`, but this does not make much difference in the dark theme.

image
image
image
image
image


parallels999:
looks great to me

PR #1827, opened 2025-12-19
Allow Symfony v8

jnoordsij:
See also https://symfony.com/releases/8.0 and https://github.com/symfony/finder/blob/8.0/CHANGELOG.md (i.e. this should effectively be no change).


barryvdh:
Thanks!

PR #1863, opened 2026-01-06
Bring back logs collector

erikn69:
This collector was removed in #1846, 
But the provider was created; I'm assuming it was removed by mistake, since the entire collector is still there.

https://github.com/barryvdh/laravel-debugbar/blob/0b6127228d80fc90e44f1dc90bfdc00d162e09e4/readme.md?plain=1#L29


barryvdh:
Wasn't really sure, I actually wanted to delete it. Don't really see the use for it, or do you? It doesnt hurt to bring it back I guess.

PR #1865, opened 2026-01-07
Update message handling in GateCollector

erikn69:
- https://github.com/php-debugbar/php-debugbar/pull/903#issuecomment-3719185545

`customizeMessageHtml` was apparently used to assign the ability as the main message instead of `array:5`. Changing the method name loses this functionality, and it also doesn't execute without `compactDumps`. Any other solutions?


barryvdh:
See https://github.com/barryvdh/laravel-debugbar/pulls

PR #1866, opened 2026-01-07
Use message context for gate and logs

barryvdh:
Based on https://github.com/php-debugbar/php-debugbar/pull/905

Screenshot 2026-01-07 at 16 13 22

### DIFF
--- a/src/DataCollector/GateCollector.php
+++ b/src/DataCollector/GateCollector.php
@@ -32,22 +32,6 @@ class GateCollector extends MessagesCollector
         });
     }
 
-    /**
-     * {@inheritDoc}
-     */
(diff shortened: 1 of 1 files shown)


erikn69:
I'm having CSS problems

In logs, The log preview displays all the text, creating a large scroll bar, and the first line is empty; that didn't happen before.
image

On gatecollector, I don't know if it's just me, but that shade of green is hard to read; before, I think it was yellow.
image

Another separate case is the timeline widget; there's a small overflow that creates a scroll bar.
image

barryvdh:
I’ll try with https://github.com/barryvdh/laravel-debugbar/pull/1864 which should avoid some conflicts

PR #1879, opened 2026-01-12
Fix explain option access in DatabaseCollectorProvider

erikn69:
Maybe just `'explain' => true/false` on config


erikn69:
Furthermore, after the last optimization, the explain function no longer runs automatically; now a button is used. In my opinion, it should always be active.

PR #1880, opened 2026-01-12
Make 'explain' option default

erikn69:
- Alternative to #1879

https://github.com/fruitcake/laravel-debugbar/pull/1879#issuecomment-3740573610
>Furthermore, after the last optimization, the explain function no longer runs automatically; now a button is used. In my opinion, it should always be active.


barryvdh:
hmm yeah maybe we just make it enabled by default?

barryvdh:
Changed default in in https://github.com/fruitcake/laravel-debugbar/commit/6980ac3d4eb53a6728ac83ef948bcede47ad0b15

PR #1896, opened 2026-01-15
Separate the debugbar from the application load(TimeCollector)

erikn69:
This is a proof of concept; it may need more work.

image

I've noticed that the debugbar can take up to a third of the application's load time. Would it be a good idea to try to differentiate the time the debugbar takes from the application's load time?


barryvdh:
Yea I guess so

PR #1943, opened 2026-01-26
Update onCacheEvent to accept multiple event types

Yahav:
This fixes https://github.com/fruitcake/laravel-debugbar/issues/1942

The reason these classes (specifically the flush-related ones) do not extend CacheEvent is that CacheEvent requires a $key property in its constructor, whereas flush operations typically affect the entire store or tags rather than a specific key. CacheFailedOver also has a different structure, focusing on the store name and the exception encountered.


barryvdh:
Thanks! Was just about to fix with https://github.com/fruitcake/laravel-debugbar/pull/1944 but merged this first.

PR #1947, opened 2026-01-28
Allow ability parameter to accept integer type

jeffersongoncalves:
In this project, we use integer values ​​in the gate.

image

### DIFF
--- a/src/DataCollector/GateCollector.php
+++ b/src/DataCollector/GateCollector.php
@@ -20,7 +20,7 @@ class GateCollector extends MessagesCollector implements Resettable
 {
     protected array $reflection = [];
 
-    public function addCheck(mixed $user, string $ability, mixed $result, array $arguments = []): void
+    public function addCheck(mixed $user, string|int $ability, mixed $result, array $arguments = []): void
     {
         $userKey = 'user';
         $userId = null;


erikn69:
$ability should not be `int` according to the docs,
you should cast those ints to strings

DOCS: https://laravel.com/docs/12.x/authorization#intercepting-gate-checks
PHPDOC Source: [Illuminate/Auth/Access/Gate.php#L428](https://github.com/laravel/framework/blob/00f0216e493867f399401397cc40347f869a7e1e/src/Illuminate/Auth/Access/Gate.php#L428)

jeffersongoncalves:
The problem is that the method signature doesn't specify a string, only the PHPDoc for the method.

Since the value in GateCollector.php is typed as a string, this caused the problem, which led to a PR request to add the int type. This is because the Laravel Gate core itself doesn't have the string value fixed in its `$ability`.

Gate->define allows string or UnitEnums: [Illuminate/Auth/Access/Gate.php#L203](https://github.com/laravel/framework/blob/00f0216e493867f399401397cc40347f869a7e1e/src/Illuminate/Auth/Access/Gate.php#L203)

barryvdh:
I don't think we need to be stricter then Laravel, we want to avoid triggering errors on anything that we can do in Laravel, even it's not totally correct

PR #1953, opened 2026-02-03
Add cursor style for resize handle in debugbar

finagin:
https://github.com/user-attachments/assets/52a4a6be-6c8d-471f-abd5-0b89bcd7a37f

### DIFF
--- a/resources/laravel-debugbar.css
+++ b/resources/laravel-debugbar.css
@@ -93,6 +93,7 @@ div.phpdebugbar-openhandler {
 }
 
 div.phpdebugbar-resize-handle {
+    cursor: row-resize;
     height: 3px;
     margin-top: -3px;
     border-bottom: 0;


erikn69:
https://github.com/php-debugbar/php-debugbar/pull/401
[php-debugbar/resources/debugbar.css#L232-L239](https://github.com/php-debugbar/php-debugbar/blob/f52f0f148a6cc32b4a16c00ce74f369b81bea378/resources/debugbar.css#L232-L239)

How is it different from the one bundled with the package?

miken32:
I was going to come here with the same suggestion. But after minimizing/maximizing the debug bar once, everything is now looking ok.

The first time I loaded a page after the update to 4.0, the `.phpdebugbar-minimized` class was applied to the debugbar even though it was open, so the cursor wasn't set properly.

finagin:
Thank you @erikn69 

I found a discussion about this issue on reddit.
This is a Safari 26.2 issue.
The following cursor options don't work: `e-resize`, `ew-resize`, `n-resize`, `ne-resize`, `nesw-resize`, `ns-resize`, `nw-resize`, `nwse-resize`, `s-resize`, `se-resize`, `sw-resize`, `w-resize`

https://www.w3schools.com/cssref/tryit.php?filename=trycss_cursor

In any case, it's worth sending this MR to the parent package, I think

erikn69:
> In any case, it's worth sending this MR to the parent package, I think

Yes, that would be better.

barryvdh:
I can't find any difference between browser support. See https://caniuse.com/mdn-css_properties_cursor_row-resize

If there is a difference there might be an error somewhere

miken32:
@barryvdh I can confirm I'm seeing the problem on the \<shudder> w3schools link above using Safari on desktop, despite what caniuse says here: https://caniuse.com/mdn-css_properties_cursor_ns-resize. `ns-resize` doesn't work, `row-resize` does.

barryvdh:
So https://github.com/php-debugbar/php-debugbar/pull/988 ?

barryvdh:
> I was going to come here with the same suggestion. But after minimizing/maximizing the debug bar once, everything is now looking ok.
> 
> The first time I loaded a page after the update to 4.0, the `.phpdebugbar-minimized` class was applied to the debugbar even though it was open, so the cursor wasn't set properly.

I think I found the issue. When no previous state was found, the bar had the minimized class, but was not minimized. That put the cursor on 'auto'. I've fixed this state in the mentioned PR.

PR #1955, opened 2026-02-03
fix: urlencode cache key

calebdw:
Hello!

@barryvdh,

The url generation fails when the cache key contains slashes or something that looks like a route parameter. We just need 
to urlencode the key to prevent this.

This fixes the following error we are getting after upgrading to v4:

```
Missing required parameter for [Route: debugbar.cache.delete] [URI: _debugbar/cache/{key}] [Missing parameter: resource].
```

Thanks!


barryvdh:
Hmm I would think that Laravel would handle that automatically, but apparently not.

barryvdh:
I'm thinking that maybe we should put the key in the body?

barryvdh:
Oh nevermind, I put it in the request URL because I want to have it signed.

PR #1957, opened 2026-02-05
chore: replace deprecated Request::get() with Request::input()

calebdw:
Hello!

The `Request::get()` method has been deprecated and is generating the following logs: `Since symfony/http-foundation 7.4: Request::get() is deprecated, use properties ->attributes, query or request directly instead. in /var/www/eaglesys/releases/255/vendor/symfony/deprecation-contracts/function.php on line 25
`

Thanks!


calebdw:
Thanks!

barryvdh:
Although I think this was from the request data collector

PR #1968, opened 2026-02-17
Cast LARAVEL_START const to float

daniser:
LARAVEL_START constant may be a string, for example when configured by PHPUnit configuration file:
```xml
<env name="DEBUGBAR_ENABLED" value="false"/>
<const name="LARAVEL_START" value="0"/>
```
This leads to following error:
image

Note that disabling Debugbar doesn't help because exception is thrown in `LaravelDebugbar` class constructor.

This PR enforces casting LARAVEL_START constant to a float.

### DIFF
--- a/src/LaravelDebugbar.php
+++ b/src/LaravelDebugbar.php
@@ -215,7 +215,7 @@ class LaravelDebugbar extends DebugBar
 
     public function booted(): void
     {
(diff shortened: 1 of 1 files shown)


barryvdh:
Okay but why is it running in tests?

daniser:
> Okay but why is it running in tests?

Good question. I actually don't want to run it in tests, that's why I set `DEBUGBAR_ENABLED` to `false`.

PR #1976, opened 2026-03-04
Add button to show query results

barryvdh:
Feature request: Run queries to see the result.

It's basically the same as Explain, only without the 'EXPLAIN' part.

We can run it as a table, or a dump object.

Tables can now be expanded (also for the explain) and truncated, but full text + selected on click


barryvdh:
Screenshot 2026-03-04 at 16 44 57

Issue #1826, opened 2025-12-13
Argument #1 ($haystack) must be of type string, array given


Prithviraj639:
# TypeError - Internal Server Error

strpos(): Argument #1 ($haystack) must be of type string, array given

PHP 8.5.0
Laravel 12.42.0
127.0.0.1:8000

## Stack Trace

0 - vendor\barryvdh\laravel-debugbar\src\DataCollector\QueryCollector.php:381
1 - vendor\barryvdh\laravel-debugbar\src\DataCollector\QueryCollector.php:381
2 - vendor\barryvdh\laravel-debugbar\src\DataCollector\QueryCollector.php:331
3 - vendor\barryvdh\laravel-debugbar\src\DataCollector\QueryCollector.php:279
4 - vendor\barryvdh\laravel-debugbar\src\DataCollector\QueryCollector.php:181
5 - vendor\barryvdh\laravel-debugbar\src\LaravelDebugbar.php:414
6 - vendor\laravel\framework\src\Illuminate\Events\Dispatcher.php:488
7 - vendor\laravel\framework\src\Illuminate\Events\Dispatcher.php:315

cyppe:
TypeError
vendor/php-debugbar/php-debugbar/src/DebugBar/DataFormatter/HasXdebugLinks.php:38
str_starts_with(): Argument #1 ($haystack) must be of type string, false given

This was in new beta4.. but maybe also in regular.

barryvdh:
First one: https://github.com/barryvdh/laravel-debugbar/pull/1845
Second one: https://github.com/php-debugbar/php-debugbar/pull/870

PR #1979, opened 2026-03-05
Cast database name to string in QueryCollector for null values

b8ne:
Ref issue #1826


barryvdh:
It's only for the str contains check right? So maybe best to keep it null and tweak the check (see commit)

PR #1989, opened 2026-03-06
Easy collector getter from `debugbar()` helper

ssw1cblarrion:
You could do: 
```php
debugbar('queries')?->addMessage('message');
```

Would something like that be unthinkable?


barryvdh:
Yeah I like it

PR #1998, opened 2026-03-09
Simplify explain option on config

erikn69:
Is there any reason why it should still be an array?


barryvdh:
It used to be because there were more options (eg which type of queries)

barryvdh:
Also needs a check in the QueryController

erikn69:
Done

barryvdh:
Yeah I guess, but feels a bit weird. Let me ponder on it a bit

erikn69:
> Yeah I guess, but feels a bit weird

Another approach made

PR #2001, opened 2026-03-20
Fix clockwork support

erikn69:
it is failing
image

### DIFF
--- a/src/Support/Clockwork/Converter.php
+++ b/src/Support/Clockwork/Converter.php
@@ -65,14 +65,12 @@ class Converter
         if (isset($data['route'])) {
             $route = $data['route'];
 
-            $controller = null;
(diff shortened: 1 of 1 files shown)


erikn69:
After the switch to the new json formatter, the message log no longer shows useful information when using arrays.
image

barryvdh:
> After the switch to the new json formatter, the message log no longer shows useful information when using arrays. image

Is that a different issue? Probably need to trigger the json dumper there. How to reproduce?

erikn69:
> Is that a different issue?

Yes. it is

The problem is that apparently before there was a preview of the JSON, now there is only `array(1)`, without any information about the content.

> How to reproduce?

`\Debugbar::log(['test' => 'test']);` and check log tab on clockwork

PR #2014, opened 2026-03-31
Add debug for failed queries

erikn69:
It displays the QueryException as an executed query but with the attached error
image

Is there a better way to display failed queries?

UPDATE:
with `array_filter` JSON becomes lighter


barryvdh:
I think I like this. I was looking into the exception handling, and currently we just add exceptions when attached to the response. I tried adding the logic there, which works but then it shows the query at the end, which might be incorrect.

But I guess the bigger case would be of the `report($e)` method, which should probably also trigger the exception handler from debugbar.

So I'm thinking we should:
 - Move the Exception handler stuff to the exception collector provider 
 - Check exception + previous exception for queries

But maybe it could make more sense to split them, I guess 🤔

barryvdh:
Updated in https://github.com/fruitcake/laravel-debugbar/pull/2015

erikn69:
Great 👍

PR #2020, opened 2026-04-03
feat: add editor links to SQL query backtrace entries

shaffe-fr:
Each backtrace frame in the queries panel now generates an `xdebug_link`, allowing clickable links to open files directly in the configured editor (phpstorm, vscode, etc.), just like the main query filename already does.

## Changes

- **QueryCollector.php**: Generate an `xdebug_link` for each backtrace entry via `getXdebugLink()`
- **resources/queries/widget.js**: Override backtrace rendering in `itemRenderer` to append an editor link icon next to each trace frame

## Implementation note

The JS override works by finding the Backtrace `<thead>` rendered by the parent `SQLQueriesWidget` and replacing the `<tbody>` content with editor-linked entries. This is a bit fragile (depends on the `Backtrace` label text), but I couldn't find a cleaner approach since the base widget doesn't expose any hook to customize the backtrace rendering. Open to suggestions if there's a better way.

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -101,6 +101,7 @@ return [
             ],
             'backtrace'         => env('DEBUGBAR_OPTIONS_DB_BACKTRACE', true),   // Use a backtrace to find the origin of the query in your files.
             'backtrace_exclude_paths' => [],   // Paths to exclude from backtrace. (in addition to defaults)
+            'backtrace_editor_links' => env('DEBUGBAR_OPTIONS_DB_BACKTRACE_EDITOR_LINKS', false), // Add editor links to backtrace entries (non-vendor files only)
             'timeline'          => env('DEBUGBAR_OPTIONS_DB_TIMELINE', false),  // Add the queries to the timeline
             'duration_background'  => env('DEBUGBAR_OPTIONS_DB_DURATION_BACKGROUND', true),   // Show shaded background on each query relative to how long it took to execute.
             'explain' => [                 // Show EXPLAIN output on queries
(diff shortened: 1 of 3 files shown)


barryvdh:
I'm a bit worried this is overkill. Should we not just try to find the most relevant item?

shaffe-fr:
I understand the concern. I've updated the PR to only generate editor links for non-vendor files, so framework internals stay as plain text.

The use case I had in mind is when the backtrace shows multiple app-level files. For example, a query triggered from a controller that goes through a service, a repository, and a data class:

```
0  app/Features/Homepage/Data/HomepageProductsData.php:31  🔗
1  app/Features/Homepage/Services/HomepageDataService.php:28  🔗
2  app/Http/Controllers/HomepageController.php:29  🔗
3  vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php:46
4  vendor/laravel/framework/src/Illuminate/Routing/Route.php:269
```

Right now, only the top-level file (the one shown next to the query) has an editor link. But when debugging, I often need to jump to the service or controller that initiated the call chain, not just the final caller. Having to manually search for the file and line number adds friction.

With this change, only the app-level frames (0, 1, 2) get editor links. Vendor frames stay untouched.

parallels999:
It's not opt-in, and I can't opt ​​out of this option.

It seems overkill to me, there is already a link to the class that runs the query

shaffe-fr:
I've reworked the PR so the feature is now opt-in via a backtrace_editor_links config option (defaults to false).
To enable it:

```
// config/debugbar.php
'db' => [
    'backtrace_editor_links' => true,
],
```
Or via env: `DEBUGBAR_OPTIONS_DB_BACKTRACE_EDITOR_LINKS=true`

PR #2027, opened 2026-04-09
Enhance storage open check with additional config

erikn69:
Fixes https://github.com/fruitcake/laravel-debugbar/pull/2026#issuecomment-4218145263


barryvdh:
Maybe we can re-use the static check there?

PR #2028, opened 2026-04-10
Avoid hash mismatch by stringify bindings

erikn69:
When using `\Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull`
Bindings like 
```
array:3 [
 0 => ""
 1 => ""
 2 => "" 
]
```
transform into
```
array:3 [
 0 => null
 1 => null
 2 => null 
]
```
Therefore, the hashes do not match, and you cannot use explain or run the query.

### DIFF
--- a/resources/queries/widget.js
+++ b/resources/queries/widget.js
@@ -96,7 +96,7 @@
             const body = {
                 connection: statement.explain.connection,
                 query: statement.explain.query,
-                bindings: statement.params,
+                bindings: JSON.stringify(statement.params),
                 hash: statement.explain.hash,
                 mode: mode,
             };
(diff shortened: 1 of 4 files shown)


barryvdh:
What if we just send the request id and the hash, like the CLI does?

erikn69:
I didn't understand it,
with request_id you would find the JSON file, and find the query with the hash?
In that case, wouldn't it be better to use UUIDs for the queries instead of hashes?
And what if the JSON file doesn't exist? If my memory serves me right, there was the `defer_datasets` configuration.

I would have to try

barryvdh:
I'm not sure if the UUID is actually faster though, Claude doesn't seem to think so.

defer_dataset would still have the ID and just load the dataset a bit later.

PR #2039, opened 2026-05-03
Fix description for `debugbar:get` command

OussamaMater:
Hey 👋🏽

While upgrading from v3 to v4 and trying out the new CLI, I noticed that the `debugbar:get` command has a slightly misleading description. It doesn't list requests, it actually inspects a single one. Seems to be a duplicate of `debugbar:find`.

I updated the description and tried to match the `debugbar:queries` command's tone, but feel free to tweak it.

Thanks for the new CLI tool, it's super helpful.


barryvdh:
Thanks!

PR #2058, opened 2026-06-18
Bring back view measuring

sysdev34-wq:
OPT-IN on  `debugbar.options.views.timeline_duration = true`

Replaces [vdauchy/laravel-debugbar-view-meter](https://github.com/vdauchy/laravel-debugbar-view-meter)

image

Currently it only shows the time of the view, but not the duration; is it possible to have it again, at least as an OPT-IN?

### DIFF
--- a/config/debugbar.php
+++ b/config/debugbar.php
@@ -118,6 +118,7 @@ return [
         ],
         'views' => [
             'timeline' => env('DEBUGBAR_OPTIONS_VIEWS_TIMELINE', true),                  // Add the views to the timeline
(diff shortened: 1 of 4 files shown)


sysdev34-wq:
Any suggestions or comments?

It can be useful for finding a view that takes too long and optimizing it.

barryvdh:
I don't know. I removed it because it felt to 'intrusive' or hacky. I'd rather not change around to much to avoid differences between live and debug versions.

sysdev34-wq:
It used to be very intrusive because it always registered the wrapper; now, the wrapper is only registered if the necessary conditions are met—if the configuration isn't activated, the native engine remains.

It is also the only way I found to get the view load times, which is helpful.

sysdev34-wq:
Thanks

PR #2061, opened 2026-07-02
Feature/add ai collector

NielsSmits1:
## Add `AiCollector` for `laravel/ai` agent runs

Adds an `AiCollector` (and provider) that shows each `laravel/ai` agent invocation in the debug bar as a single entry: prompt, instructions, model/provider, tool calls and token usage. Built in the same style as the existing `PennantCollector`.

Screenshot 2026-07-02 at 11 09 28

### How it works

- **One entry per invocation.** `ToolInvoked` fires during a run and `AgentPrompted` at the end; tool calls are buffered by `invocationId` and folded into the entry recorded on `AgentPrompted` (`AgentStreamed` is handled identically).
- **Guarded, zero overhead when unused.** The provider only registers when `Laravel\Ai\AiManager` exists — apps without `laravel/ai` are unaffected, just like `PennantCollectorProvider` guards on `FeatureManager`.
- **No new front-end assets.** Reuses the existing `JsonVariableListWidget` / `HtmlVariableListWidget` / `VariableListWidget`, chosen the same way as `SessionCollector` / `RouteCollector`.
- **Kept lean.** Prompt/response/tool bodies are truncated at 20k chars; attachments are counted only.

### Configuration

```php
'collectors' => [
    'ai' => env('DEBUGBAR_COLLECTORS_AI', true),
],
'options' => [
    'ai' => [
        'values' => env('DEBUGBAR_OPTIONS_AI_VALUES', true), // Collect prompt/response/tool bodies
    ],
],
```

Set `'values' => false` for metadata only (agent, model, provider, usage, attachment count), handy when prompts are sensitive or large. `laravel/ai` is added to `require-dev` only — never a hard `require`.

### Tests

`AiCollectorTest` (5 tests): provider registration, folding tool calls into the matching invocation, per-invocation buffer consumption, the `values` toggle, and truncation.

### Caveat

Debugbar captures data per HTTP request, so you only see AI runs for synchronous calls (`QUEUE_CONNECTION=sync` or inline); for queued jobs a Telescope watcher fits better.


olivermbs:
Thanks for adding this! I’m hitting a compatibility issue with `laravel/ai` **1.x-dev** (commit `c1d8aeca480ffe107fa79b92c30946efbbd2d94d`) and Debugbar **v4.4.3**.

`AiCollector::recordAgentPrompted()` reads `$usage->promptTokens` and `$usage->completionTokens`, but the SDK now exposes `$usage->inputTokens` and `$usage->outputTokens`.

This throws `Undefined property: Laravel\Ai\Responses\Data\TextUsage::$promptTokens` after a successful agent response, causing the application request to fail.

Reproduced with Debugbar enabled and a faked agent response. Disabling the AI collector resolves it.

Could the collector support the new usage fields while retaining compatibility with earlier SDK versions? Happy to open a separate issue if preferred.

PR #2064, opened 2026-07-04
Resolve component name instead of verifying snapshot

### LINKED ISSUE #2063: Debugbar trigger Livewire snapshot verfication
When developing a feature related to `PersistentMiddleware` (https://github.com/livewire/livewire/pull/10382), I came across something unsual about persistent middleware applied twice. 

### PersistentMiddleware
```php
function boot()
{
    // triggered twice
    on('snapshot-verified', function ($snapshot) {
        // Only apply middleware to requests hitting the Livewire update endpoint, and not any fake requests such as a test.
        if (! app(HandleRequests::class)->isLivewireRoute()) return;

        $this->extractPathAndMethodFromSnapshot($snapshot);

        $this->applyPersistentMiddleware();
    });
}
```

At first I thought that my feature trigger it twice, however after some digging, I found the root cause 

### File `/src/DataCollector/RouteCollector.php`
```php
protected function getRouteInformation(mixed $route): array
{
    ...

    [$component] = app(HandleComponents::class)->fromSnapshot($snapshot);

    ...
}
```

This cause Livewire event handler fired twice; from the application itself and from debugbar.

```php
public function fromSnapshot($snapshot)
{
    Checksum::verify($snapshot);

    trigger('snapshot-verified', $snapshot); // here

    ...
}
```
Debugbar is really helpfull for me so is there a way to prevent this while in development? It really cause some issue if its related middleware

### Suggestion

This is a snippet from Livewire how to resolve component class name

```php
$uses = $action['uses'] ?? false;

if (! $uses) return false;

if (is_string($uses)) {
    $class = str($uses)->before('@')->toString();
    $method = str($uses)->after('@')->toString();

    // Case 1: Direct component class usage (Route::get('/path', Component::class))
    if (is_subclass_of($class, \Livewire\Component::class) && $method === '__invoke') {
        return $class;
    }

    // Case 2: Route::livewire macro usage (via LivewirePageController)
    if (str($uses)->contains(LivewirePageController::class)) {
        $component = $action['livewire_component'] ?? null;

        if (! $component) return false;

        return app('livewire.factory')->resolveComponentClass($component);
    }
}
```

barryvdh:
Fixes https://github.com/fruitcake/laravel-debugbar/issues/2063

### DIFF
--- a/src/DataCollector/RequestCollector.php
+++ b/src/DataCollector/RequestCollector.php
@@ -13,7 +13,6 @@ use Illuminate\Support\Facades\Artisan;
 use Illuminate\Support\Str;
 use Laravel\Telescope\IncomingEntry;
 use Laravel\Telescope\Telescope;
-use Livewire\Mechanisms\HandleComponents\HandleComponents;
 use Symfony\Component\Console\Input\ArgvInput;
 
 class RequestCollector extends SymfonyRequestCollector implements DataCollectorInterface, Renderable
--- a/src/DataCollector/RequestCollector.php
+++ b/src/DataCollector/RequestCollector.php
@@ -157,7 +157,7 @@ class RequestCollector extends SymfonyRequestCollector implements DataCollectorI
         $uses = $action['uses'] ?? null;
         $controller = is_string($action['controller'] ?? null) ? $action['controller'] : '';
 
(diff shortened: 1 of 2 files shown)


ghabriel25:
@barryvdh Thanks! still working on other PR so couldnt make this fix

PR #2083, opened 2026-09-01
Stringify models passed as Gate context arguments, not just arguments[0]

### LINKED ISSUE #2081: GateCollector captures a live Eloquent model when authorizing resource creation using [ClassName, model] argument shape
### Description

`GateCollector::addCheck()` only guards against `$arguments[0]` being an `Illuminate\Database\Eloquent\Model`:

```
// src/DataCollector/GateCollector.php
if (isset($arguments[0])) {
    if ($arguments[0] instanceof Model) {
        $model = $arguments[0];
        $target = get_class($model) . '(' . $model->getKeyName() . '=' . $model->getKey() . ')';
        $arguments[0] = $target;
    } elseif (is_string($arguments[0])) {
        $target = $arguments[0];
    }
}

$this->addMessage("{ability} {target}", $label, [
    ...
    'arguments' => $arguments,
]);
```

This misses Laravel's documented convention for authorizing the creation of a not-yet-existing resource:

`Gate::allows('create', [Post::class, $category]);`

Here `arguments[0]` is the policy's class name (a string), and `arguments[1]` is the actual model providing context to the `create()` ability. Since the code only inspects index 0, it falls into the `elseif (is_string(...))` branch and never sanitizes `arguments[1]` - the real model is stored unmodified, as a live reference, in the collector's message array.

### Why this matters

A live Eloquent model retained by the collector isn't just inert data, it can carry Eloquent's internal state along with it. In particular, with `Model::automaticallyEagerLoadRelationships()` enabled, every model carries a `relationAutoloadCallback` / `relationAutoloadContext` pair that references back to the entire collection it was hydrated with (this is intentional, it's what lets Laravel batch-load relations across a whole result set). Capturing one row's model via this bug means capturing a live reference to the whole page's result set, once per Gate check.

On a paginated index page (e.g. 100 rows) with a per-row `create` authorization check (a completely ordinary pattern, e.g. "can I create a quote for this customer" rendered as a disabled/enabled button per row), this means ~100–200 Gate messages each holding a reference to the full 100-row collection. Depending on what else is loaded on those models and how the toolbar renders the captured arguments, this can consume very large amounts of memory building the debug toolbar itself, in our case enough to exhaust the PHP memory limit and take down the request, on a page that (confirmed via profiling) legitimately uses only a few MB without Debugbar's gate collector enabled.

### Reproduction

```
// Any policy authorizing creation of a resource, using context from another model:
class PostPolicy
{
    public function create(User $user, Category $category): Response
    {
        return $category->owner_id === $user->id
            ? Response::allow()
            : Response::deny();
    }
}

// In a loop over a paginated collection, e.g. a Blade view:
@foreach ($categories as $category)
    {{ Auth::user()->cannot('create', [Post::class, $category]) }}
@endforeach
```

With the `gate` collector enabled, open the Debugbar Gate tab and inspect any `create` check's arguments.

`arguments[1]` is the live `$category` model, not a stringified representation, and (with automaticallyEagerLoadRelationships() on) its relationAutoloadContext property references the entire paginated collection.

### Suggested fix

Scan the whole `$arguments` array for `Model` instances, not just index 0:

```
foreach ($arguments as $i => $argument) {
    if ($argument instanceof Model) {
        $stringified = get_class($argument) . '(' . $argument->getKeyName() . '=' . $argument->getKey() . ')';
        if ($i === 0) {
            $target = $stringified;
        }
        $arguments[$i] = $stringified;
    } elseif ($i === 0 && is_string($argument)) {
        $target = $argument;
    }
}
```

AdilAzhariOmsan:
Closes #2081.

### Problem

`GateCollector::addCheck()` only stringifies `$arguments[0]` when it is a `Model`. Laravel's documented shape for authorizing creation of a not-yet-existing resource passes the model at a **later** index:

```php
Gate::allows('create', [Post::class, $category]);
```

Here `$arguments[0]` is the policy class name (a string) and `$arguments[1]` is the context model. The old code inspects index 0 only, so it takes the `is_string()` branch and `$category` is stored **unmodified** — a live Eloquent reference — in the collector's message array.

With `Model::automaticallyEagerLoadRelationships()` enabled, every hydrated model carries a `relationAutoloadCallback` / `relationAutoloadContext` pair that references the whole collection it was loaded with. On a paginated index page that runs a per-row `create` check (an ordinary "can I create X for this row" enabled/disabled-button pattern), the collector then retains roughly one reference to the entire page's result set per check — enough, on a large page, to exhaust the PHP memory limit while the toolbar itself is being built.

### Fix

Scan the whole `$arguments` array for `Model` instances and stringify each one. The index-0 `$target` logic (string or model → `$target`) is unchanged.

### Tests

Added `GateCollectorTest::testItStringifiesModelsPassedAsContextArguments` — authorizes with the `[User::class, $person]` shape and asserts `arguments[1]` is stored as `Person(id=2)`, not a live model. Existing `testItCollectsGateChecks` still passes.

`vendor/bin/pint --test` and `vendor/bin/phpstan analyse` on the changed file both clean.


barryvdh:
what is the difference with https://github.com/fruitcake/laravel-debugbar/pull/2082

AdilAzhariOmsan:
#2082 landed the same fix while this was in review, and its test covers the same [Class::class, $model] case. I should have checked the issue's linked PRs before opening this. Closing in favour of #2082.

PR #2086, opened 2026-09-22
Fix AI collector compatibility with Laravel AI 1.x token usage

olivermbs:
A successful Laravel AI response can fail the application request when Debugbar collects it: `AiCollector::recordAgentPrompted()` reads `promptTokens` / `completionTokens`, but Laravel AI 1.x exposes `inputTokens` / `outputTokens` on `TextUsage`. The exception occurs after the AI response succeeds, so debug instrumentation breaks an otherwise successful request. Both `AgentPrompted` and `AgentStreamed` reach this code.

Reported with Debugbar v4.4.3 and `laravel/ai` `1.x-dev` at `c1d8aeca480ffe107fa79b92c30946efbbd2d94d`. The collector was introduced in #2061.

Reproduction:

1. Enable Debugbar and its AI collector with that SDK revision installed.
2. Complete an agent prompt or consume an agent stream.
3. When the completion event is dispatched, collection throws `Undefined property: Laravel\Ai\Responses\Data\TextUsage::$promptTokens`.

The added event-dispatch tests reproduce this without a network request or API credentials. Before the fix, both event paths fail with that exception on the reported SDK revision.

The fix reuses the usage array already collected for display, preferring `input_tokens` / `output_tokens` and falling back to `prompt_tokens` / `completion_tokens`. Null coalescing preserves valid zero values. The detailed usage payload remains unchanged.

Regression coverage uses the installed SDK's real usage classes and a real `StreamedAgentResponse`, dispatching both completion events through the registered collector listeners. It checks nonzero counts, zero input, zero output, both zero, the detailed usage payload, and response text. Existing fixtures also supply tool timing when required by 1.x.

The development constraint now permits `^0.8|^1.0`. A focused CI matrix explicitly installs each version range and runs `AiCollectorTest`, ensuring `prefer-stable` cannot leave 1.x untested. Laravel AI remains a development-only dependency.

Validation on PHP 8.4.23 / Laravel 13.32.0:

- `composer test -- --filter=AiCollectorTest`: 13 tests, 61 assertions passed on AI v0.8.1, the reported 1.x commit, and current 1.x commit `2b6daad566dadd237d2548907a57f3ec7da260e6`.
- `composer analyse -- --no-progress --debug`: passed with all three SDK revisions.
- `composer check-style`, `composer validate --no-check-lock`, workflow YAML parsing, and `git diff --check`: passed.
- Full `composer test`: 294 tests, 652 assertions; six cache URL failures and two missing-Chromedriver errors. The same six failures and two errors reproduce with the unchanged upstream collector and tests on AI v0.8.1 (286 tests, 620 assertions before the eight added cases).
- PHPUnit reports three existing doc-comment metadata deprecations in `RouteCollectorTest`.


barryvdh:
Thanks! test failures were unrelated. Follwed up with matrix in https://github.com/fruitcake/laravel-debugbar/pull/2090

